PR #4, opened 2019-04-12
[KEYCLOAK-9344] - REST API Guideline

pedroigor:
The task list below will be used to track changes agreed during discussions:

- [x] Include `extension_id` to the extension resource path format
- [x] Include HTTP Method `PATH` to the list of HTTP verbs 
- [x] Extensions to the apis root path. E.g: `/apis/extensions/{extension_id}' 
- [ ] Add a recommendation section
- [x] Include 'rate limiting' as s security recommendation 
- [x] Expand level of maturity section instead of linking to external resource
- [x] Resource paths should be prefixed with the realm
- [x] Provide more details about Controller Resources
- [x] Remove `GET` from the supported list of methods that can be used to perform actions and use POST instead

### DIFF
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -65,10 +65,10 @@ By respecting this principle, we expect developers to provide more concise and c
 In Keycloak, we choose to stick with the following convention for resource paths:
 
 ```
-/realms/{realm}/apis/{API_GROUP}/{version}
+/{realm}/apis/{API_GROUP}/{version}
 ```
 
-Keycloak APIs usually operate in the context of a realm, the resource path should be prefixed with `/realms/{realm}` where `realm` maps to the the name of an existing realm.
+Keycloak APIs usually operate in the context of a realm, the resource path should be prefixed with `/{realm}` where `realm` maps to the the name of an existing realm.
 
 In Keycloak, a realm represents a tenant from where all the configuration is done. By operating in the context of a specific realm, APIs should always consider the realm specified in the path when performing any operation.
 
@@ -81,7 +81,7 @@ By including the group in the resource path, we aim to provide a clear and consi
 Here is an example of a resource path related with the group of functionalities provided by the Account Management REST API:
 
 ```
-/realms/{realm}/apis/accounts/v1/{user_id}
+/{realm}/apis/accounts/v1/{user_id}
 ```
 
 The path refers to the `accounts` group and any resource/operation related with user account management (in this case for user with id `user_id`) can be accessed from that path.
@@ -94,12 +94,18 @@ Group names should be in lowercase and a single word name.
 Under some circunstances, the group can be a compound name. This is the case when APIs are logically related with a common group. For instance, APIs related with the Keycloak Admin API are grouped under:
 
 ```
-/realms/{realm}/apis/admin/users/v1/{user_id}
+/{realm}/apis/admin/users/v1/{user_id}
 ```
 
 The usage of compound names in groups should consider whether or not makes sense to group the API into a logical group. Taking again the Admin API example, the reasoning for using a compound name is to make easier to restrict access to administration related APIs using a Web Application Firewall.
 
 In general, try to stick with single word names for groups.
+
+##### Reserved Group Names
+
+When choosing a group name you should avoid using reserver names such as `apis` or names that conflict fully or partially to any existing group.
+
+As a recommendation for users of Keycloak, realm names should also take these considerations into account otherwise they may end up with a very odd resource path definition.
 
 #### Versioning
 
@@ -110,7 +116,7 @@ The `version` refers to the stability and the state of an API. The version does 
 Here is an example of a resource path related with the group of functionalities provided by the Account Management REST API:
 
 ```
-/realms/{realm}/apis/accounts/v1/{user_id}
+/{realm}/apis/accounts/v1/{user_id}
 ```
 
 From the example above we can infer that the path refers to a specific `v1` version with the possibility to change the version and: 
@@ -126,7 +132,7 @@ From the example above we can infer that the path refers to a specific `v1` vers
 To better accommodate and separate the core API from extensions built on top of the API, developers should consider the following path format for extensions:
 
 ```
-/realms/{realm}/apis/{API_GROUP}/extensions/{extension_id}/{version}
+/{realm}/apis/{API_GROUP}/extensions/{extension_id}/{version}
 ```
 
 This separation aims to avoid confusion when consuming stable and official Keycloak APIs so that clients can differentiate if they are using some API provided by a third-party with no support from Keycloak team.
@@ -134,7 +140,7 @@ This separation aims to avoid confusion when consuming stable and official Keycl
 When the extension does not fit into any of the existing API groups, it should be possible to use a path with the following format:
 
 ```
-/realms/{realm}/apis/extensions/{extension_id}/{version}
+/{realm}/apis/extensions/{extension_id}/{version}
 ```
 
 #### Technology Preview
@@ -142,7 +148,7 @@ When the extension does not fit into any of the existing API groups, it should b
 To better accommodate and separate the core API from technology preview features, developers should use a path format as follows:
 
 ```
-/realms/{realm}/apis/{API_GROUP}/{version}/preview/{preview_id}
+/{realm}/apis/{API_GROUP}/{version}/preview/{preview_id}
 ```
 
 Technology preview APIs are related with features that have great potential to become fully supported. They provide the latest work in some area of interest and are offered with the intent to gather feedback from the community about new functionalities. However, such functionalities can be remove at any moment and break existing clients consuming the API.
@@ -150,7 +156,7 @@ Technology preview APIs are related with features that have great potential to b
 When the API does not fit into any of the existing API groups, it should be possible to use a path with the following format:
 
 ```
-/realms/{realm}/apis/preview/{preview_id}
+/{realm}/apis/preview/{preview_id}
 ```
 
 #### Store Resources
@@ -158,7 +164,7 @@ When the API does not fit into any of the existing API groups, it should be poss
 A resource store provides operations for a given resource. For example:
 
 ```
-/realms/{realm}/apis/accounts/v1/1234/applications
+/{realm}/apis/accounts/v1/1234/applications
 ```
 
 The example above is representing a resource store path from where we can manage the `applications` for a user account with ID `1234`. Based on this path, clients can perform different operations based on the HTTP verbs or any controller resource defined to the store. Usually this type of resource is associated with CRUD operations for a specific resource.
@@ -171,7 +177,7 @@ Under some circumstances, the HTTP verbs are not enough to represent an operatio
 the resource path may indicate a specific action to be taken when accessed.
 
 ```
-POST /realms/{realm}/accounts/v1/disable HTTP/1.1
+POST /{realm}/accounts/v1/disable HTTP/1.1
 Host: keycloak.org
 ```
 
@@ -280,7 +286,7 @@ to the resource that was created:
 ```
 HTTP/1.1 201 Created
 Content-Length: 0
-Location: http://mykeycloak.com/realms/{realms}/apis/users/{version}/1234
+Location: http://mykeycloak.com/{realms}/apis/users/{version}/1234
 ```
 
 ##### Expected Status Codes
@@ -424,5 +430,5 @@ The following example demonstrates how a Users API would look and how CRUD opera
 
 |Resource Path|GET|POST|PUT|DELETE|
 |---|---|---|---|---|
-|/realms/{realm}/apis/users/v1|List all users in the realm `acme`|Creates an user in the realm `acme`|Not Supported|Not Supported|
-|/realms/{realm}/apis/users/v1/123|Gets a representation for user with id `123`|Not Supported|Updates the user with id `123`|Deletes the user with id `123`|
+|/{realm}/apis/admin/users/v1|List all users in the realm `acme`|Creates an user in the realm `acme`|Not Supported|Not Supported|
+|/{realm}/apis/admin/users/v1/123|Gets a representation for user with id `123`|Not Supported|Updates the user with id `123`|Deletes the user with id `123`|
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -153,7 +153,7 @@ To better accommodate and separate the core API from technology preview features
 /{realm}/apis/{API_GROUP}/{version}/preview/{preview_id}
 ```
 
-Technology preview APIs are related with features that have great potential to become fully supported. They provide the latest work in some area of interest and are offered with the intent to gather feedback from the community about new functionalities. However, such functionalities can be remove at any moment and break existing clients consuming the API.
+Technology preview APIs are related with features that have great potential to become fully supported. They provide the latest work in some area of interest and are offered with the intent to gather feedback from the community about new functionalities. However, such functionalities can be removed at any moment and break existing clients consuming the API.
 
 When the API does not fit into any of the existing API groups, it should be possible to use a path with the following format:
 
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -425,7 +425,8 @@ TODO
 
 ### Rate Lmiting
 
-TODO
+Keycloak does not provide rate limiting capabilities and for such it is necessary to use a third-party tool such as a API Manager.
+
 
 ### Documentation
 
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -56,7 +56,9 @@ This section defines the general principles that should be considered by API dev
 
 Based on [Richardson Maturity Model](https://martinfowler.com/articles/richardsonMaturityModel.html), developers MUST adhere to the maturity Level 2, where interactions with the API endpoints are based on the usage of HTTP Verbs and status codes to communicate status and errors.
 
-With this principle in mind, APIs should leverage the [HTTP Verbs](#http-verbs) defined in this guideline and respect their meaning, as well as use [HTTP Status Codes](#http-status-codes) in responses to communicate the success and failures when processing a request. For more details about the supported HTTP Verbs and Status Codes, see their respective sections in this document.
+With this principle in mind, APIs should leverage the [HTTP Verbs](#http-verbs) defined in this guideline and respect their meaning, idempotency as well as use [HTTP Status Codes](#http-status-codes) in responses to communicate the success and failures when processing a request. For more details about the supported HTTP Verbs and Status Codes, see their respective sections in this document.
+
+By respecting this principle, we expect developers to provide more concise and consistent APIs with a uniform interface. 
 
 ### Resource Path
 
@@ -67,6 +69,8 @@ In Keycloak, we choose to stick with the following convention for resource paths
 ```
 
 Keycloak APIs usually operate in the context of a realm, the resource path should be prefixed with `/realms/{realm}` where `realm` maps to the the name of an existing realm.
+
+In Keycloak, a realm represents a tenant from where all the configuration is done. By operating in the context of a specific realm, APIs should always consider the realm specified in the path when performing any operation.
 
 #### Group
 
@@ -84,6 +88,18 @@ The path refers to the `accounts` group and any resource/operation related with 
 Note that by grouping functionalities together and in conjunction with the versioning schema, we can keep the root path and also the semantics of the API without impact the adaptability for new changes or improvements to the API.
 
 Group names should be in lowercase and a single word name.
+
+##### Compound Group Names
+
+Under some circunstances, the group can be a compound name. This is the case when APIs are logically related with a common group. For instance, APIs related with the Keycloak Admin API are grouped under:
+
+```
+/realms/{realm}/apis/admin/users/v1/{user_id}
+```
+
+The usage of compound names in groups should consider whether or not makes sense to group the API into a logical group. Taking again the Admin API example, the reasoning for using a compound name is to make easier to restrict access to administration related APIs using a Web Application Firewall.
+
+In general, try to stick with single word names for groups.
 
 #### Versioning
 
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -419,11 +419,11 @@ TODO
 
 TODO
 
-### Concurrency Control and Concurrency
+### Concurrency Control and Consistency
 
 TODO
 
-### Rate Lmiting
+### Rate Limiting
 
 Keycloak does not provide rate limiting capabilities and for such it is necessary to use a third-party tool such as a API Manager.
 
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -45,6 +45,8 @@ Although the guideline is based on the REST architectural style, it does not try
     - [Error Handling](#error-handling)
     - [Filtering](#filtering)
     - [Pagination](#pagination)
+    - [Concurrency Control and Consistency](#concurrency-control-and-concurrency)
+    - [Rate Limiting](#rate-lmiting)
     - [Documentation](#documentation)
   - [Examples](#examples)
 
@@ -414,6 +416,14 @@ When processing error responses, developers should look the error codes being us
 TODO
 
 ### Pagination
+
+TODO
+
+### Concurrency Control and Concurrency
+
+TODO
+
+### Rate Lmiting
 
 TODO
 
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -267,7 +267,6 @@ It is expected that APIs responding to HTTP `GET` requests respect the following
 |Status Codes|
 |---|
 |200|
-|204|
 |307|
 |400|
 |401|
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -411,22 +411,54 @@ consider existing error codes before creating new ones.
 
 When processing error responses, developers should look the error codes being used by other APIs and see if they can be reused.
 
-### Filtering
-
-TODO
-
 ### Pagination
+
+When dealing with large data sets, [store resources](#store-resources) should consider paginating the results when fetching resource sets.
+
+In Keycloak, pagination is based on two request parameters:
+
+* `first`
+
+    The position of the first result to retrieve
+
+* `max`
+    
+    The maximum number of entries to retrieve
+
+These parameters should be sent along with a `GET` request as follows:
+
+```
+/{realm}/apis/admin/users/v1?first=1&max=20
+```
+
+The example above shows how to request the first page (entries from 1 to 20).
+
+As a result, the server should respond as follows:
+
+```
+HTTP/1.1 200 OK
+Connection: keep-alive
+Cache-Control: no-cache
+Content-Type: application/json
+Content-Length: 961
+Date: Wed, 22 May 2019 13:44:57 GMT
+
+Link: <http://{host}:{port}/auth/{realm}/apis/admin/users/v1?first=20&max=20>; rel="next"
+```
+
+As defined by [RFC-5988](https://tools.ietf.org/html/rfc5988), responses from the server should contain a `Link` header indicating whether or not there is a next page and, if there are more pages, a link that can be followed to fetch the next page. If the `Link` header is not present, the client can assume that there are no more pages to fetch.
+
+### Rate Limiting
+
+Keycloak does not provide rate limiting capabilities and for such it is necessary to use a third-party tool such as a API Manager.
+
+### Filtering
 
 TODO
 
 ### Concurrency Control and Consistency
 
 TODO
-
-### Rate Limiting
-
-Keycloak does not provide rate limiting capabilities and for such it is necessary to use a third-party tool such as a API Manager.
-
 
 ### Documentation
 
--- a/design/rest-api-guideline.md
+++ b/design/rest-api-guideline.md
@@ -1,0 +1,412 @@
+# REST API Guideline
+
+* **Status**: Draft
+* **Author**: keycloak-team
+* **JIRA**: [KEYCLOAK-9344](https://issues.jboss.org/browse/KEYCLOAK-9344)
+
+## Abstract
+
+The Keycloak REST API Guideline provides a set of design principles and practices that should be considered by developers when designing, implementing and exposing a RESTful API. To provide the best experience for developers consuming Keycloak REST APIs, it's important API developers to follow to the principles and practices defined herein, making these APIs intuitive and easy to use.
+
+The goals of the guideline are:
+
+* Define the [overall principles](#principles) that should be taken into account when designing REST APIs for Keycloak
+
+* Define common practices and patterns for all Keycloak REST APIs such as:
+    * [Filtering](#filtering)
+    * [Pagination](#pagination)
+    * [Error Handling](#error-handling)
+    * [Documentation](#documentation)
+
+* Define common practices for [documentation](#documentation)
+
+Although the guideline is based on the REST architectural style, it does not try to strictly adhere to RESTful constraints. The principles covered by the guideline are based on practical experiences in the development of REST APIs in Keycloak while still conforming with most of the constraints of REST architectures.
+
+**Table of Contents**
+
+
+  - [Principles](#principles)
+    - [Level of Service Maturity](#level-of-service-maturity)
+    - [Resource Path](#resource-path)
+      - [Group](#group)
+      - [Versioning](#versioning)
+      - [Extensions](#extensions)
+      - [Technology Preview](#technology-preview)
+      - [Store Resources](#store-resources)
+      - [Controller Resources](#controller-resources)
+    - [Resource Representation](#resource-representation)
+    - [HTTP Status Codes](#http-status-codes)
+    - [HTTP Verbs](#http-verbs)
+      - [GET](#get)
+      - [POST](#post)
+      - [PUT](#put)
+      - [DELETE](#delete)
+  - [Practices](#practices)
+    - [Error Handling](#error-handling)
+    - [Filtering](#filtering)
+    - [Pagination](#pagination)
+    - [Documentation](#documentation)
+  - [Examples](#examples)
+
+## Principles
+
+This section defines the general principles that should be considered by API developers, regardless of the practices defined through this document.
+
+### Level of Service Maturity
+
+Based on [Richardson Maturity Model](https://martinfowler.com/articles/richardsonMaturityModel.html), developers MUST adhere to the maturity Level 2, where interactions with the API endpoints are based on the usage of HTTP Verbs and status codes to communicate status and errors.
+
+With this principle in mind, APIs should leverage the [HTTP Verbs](#http-verbs) defined in this guideline and respect their meaning, as well as use [HTTP Status Codes](#http-status-codes) in responses to communicate the success and failures when processing a request. For more details about the supported HTTP Verbs and Status Codes, see their respective sections in this document.
+
+### Resource Path
+
+In Keycloak, we choose to stick with the following convention for resource paths:
+
+```
+/realms/{realm}/apis/{API_GROUP}/{version}
+```
+
+Keycloak APIs usually operate in the context of a realm, the resource path should be prefixed with `/realms/{realm}` where `realm` maps to the the name of an existing realm.
+
+#### Group
+
+The `API_GROUP` refers to a set of one or more functionality that can be grouped together. The endpoints associated with a group can manage different resources as far as they are strictly related with the functionality represented by the group.
+
+By including the group in the resource path, we aim to provide a clear and consistency view of the functionalities exposed by the different APIs, with more control over the groups of functionality (in terms of functional and non-functional aspects) as well as make more clear how we communicate the functional structure of these APIs through the documentation.
+
+Here is an example of a resource path related with the group of functionalities provided by the Account Management REST API:
+
+```
+/realms/{realm}/apis/accounts/v1/{user_id}
+```
+
+The path refers to the `accounts` group and any resource/operation related with user account management (in this case for user with id `user_id`) can be accessed from that path.
+Note that by grouping functionalities together and in conjunction with the versioning schema, we can keep the root path and also the semantics of the API without impact the adaptability for new changes or improvements to the API.
+
+Group names should be in lowercase and a single word name.
+
+#### Versioning
+
+Versioning helps APIs to adapt to changes to their fields or restructure resource representations. In addition to that, versioning can help developers to offer non-stable, technology preview features as well as control the end-of-life of APIs.
+
+The `version` refers to the stability and the state of an API. The version does not necessarily refers to a Keycloak release version number so that it can be mapped to different Keycloak release versions.
+
+Here is an example of a resource path related with the group of functionalities provided by the Account Management REST API:
+
+```
+/realms/{realm}/apis/accounts/v1/{user_id}
+```
+
+From the example above we can infer that the path refers to a specific `v1` version with the possibility to change the version and: 
+
+* completely change the API
+* do breaking changes to parts of the API
+* remove functionalities or specific parts of the API
+* provide non-stable or technology preview features
+* the different APIs can evolve separately
+
+#### Extensions
+
+To better accommodate and separate the core API from extensions built on top of the API, developers should consider the following path format for extensions:
+
+```
+/realms/{realm}/apis/{API_GROUP}/extensions/{extension_id}/{version}
+```
+
+This separation aims to avoid confusion when consuming stable and official Keycloak APIs so that clients can differentiate if they are using some API provided by a third-party with no support from Keycloak team.
+
+When the extension does not fit into any of the existing API groups, it should be possible to use a path with the following format:
+
+```
+/realms/{realm}/apis/extensions/{extension_id}/{version}
+```
+
+#### Technology Preview
+
+To better accommodate and separate the core API from technology preview features, developers should use a path format as follows:
+
+```
+/realms/{realm}/apis/{API_GROUP}/{version}/preview/{preview_id}
+```
+
+Technology preview APIs are related with features that have great potential to become fully supported. They provide the latest work in some area of interest and are offered with the intent to gather feedback from the community about new functionalities. However, such functionalities can be remove at any moment and break existing clients consuming the API.
+
+When the API does not fit into any of the existing API groups, it should be possible to use a path with the following format:
+
+```
+/realms/{realm}/apis/preview/{preview_id}
+```
+
+#### Store Resources
+
+A resource store provides operations for a given resource. For example:
+
+```
+/realms/{realm}/apis/accounts/v1/1234/applications
+```
+
+The example above is representing a resource store path from where we can manage the `applications` for a user account with ID `1234`. Based on this path, clients can perform different operations based on the HTTP verbs or any controller resource defined to the store. Usually this type of resource is associated with CRUD operations for a specific resource.
+
+When a resource fits under this category, the name should be a plural noun.
+
+#### Controller Resources
+
+Under some circumstances, the HTTP verbs are not enough to represent an operation that can be performed in a resource. In this case,
+the resource path may indicate a specific action to be taken when accessed.
+
+```
+POST /realms/{realm}/accounts/v1/disable HTTP/1.1
+Host: keycloak.org
+```
+
+The example above shows the usage of `disable` to operate accounts in order to disable them.
+
+When a resource fits under this category, the name should be a verb, using camel-case if necessary.
+
+##### HTTP Verbs on Controller Resources
+
+Controller resources should use the HTTP `POST` method to perform operations on a resource. There are no restrictions on the request body as long as the [resource representation](#resource-representation) is respected if the action requires a request body.
+
+### Resource Representation
+
+The standard format for resource representation is JSON. There is no rule for schemas or the semantics used by a resource representation, as long as they follow the JSON syntax properly.
+
+However, it is important to taken into account the following principles when defining resource representations:
+
+* Keep it simple for clients, so that a representation can be easily built and processed when making requests/responses to the API
+* Be aware of extensive representations as they can affect overall performance and usability, if possible review your API to see if resources can be decomposed into smaller ones
+* Avoid deeply nested JSON objects
+* Avoid using generic resource representations
+
+### Http Status Codes
+
+The tables below defines the HTTP status code that clients should expect from APIs.
+
+This document also defines which codes are expected to be used for each HTTP verb, please take a look at [HTTP Verbs](#http-verbs) for more details.
+
+Clients should consider the status codes herein defined in order to properly handle responses from APIs, recover from errors, and 
+understand the reasoning behind the API responses.
+
+#### Success Codes
+
+|Code|Description|
+|---|---|
+|200|Indicates the request completed successfully regardless the size of the response body|
+|201|Indicates the request completed successfully and a resource was created|
+|204|Indicates the request completed successfully and the server did not provide a response body|
+
+#### Redirection Codes
+
+|Code|Description|
+|---|---|
+|307|Indicates that clients should resubmit the request to another location|
+
+#### Client Side Error Codes
+
+|Code|Description|
+|---|---|
+|400|Indicates that the request is invalid, usually related with the validation of the payload|
+|401|Indicates that clients should provide authorization or the provided authorization is invalid|
+|403|Indicates that the authorization provided by the client is not enough to access the resource|
+|404|Indicates that the requested resource does not exist|
+|405|Indicates that the method chosen by the client to access a resource is not supported|
+|409|Indicates that the resource the client is trying to create already exists or some conflict when processing the request|
+|415|Indicates that the requested media type is not supported|
+
+#### Server Side Error Codes
+
+|Code|Description|
+|---|---|
+|500|Indicates that the server could not fulfill the request due to some unexpected error|
+
+### Http Verbs
+
+Based on the REST architectural style, HTTP Verbs should be used as the means to communicate to APIs the operations to be performed
+on their resources. While HTTP verbs cover by themselves the most common operations you might want to look the [Controller Resource](#controller-resources) as an alternative on how to overcome any limitation imposed by the usage of HTTP verbs herein defined.
+
+The table below defines the HTTP verbs you should consider when exposing operations through your API and the circumstances where they apply.
+
+|Verb|Description|Idempotent|
+|---|---|---|
+|GET|Requests the current state of a resource|Yes|
+|POST|Creates a new resource or submit a command|No|
+|PUT|Replaces a resource state|Yes|
+|PATCH|Applies partial updates to a resource|Yes|
+|DELETE|Removes a resource|Yes|
+
+#### GET
+
+HTTP `GET` should be used to obtain current state of a resource. It is usually related with operations on resources that return either a list or a single representation, or start some action that does not change the resource state.
+
+##### Expected Status Codes
+
+It is expected that APIs responding to HTTP `GET` requests respect the following HTTP status codes to communicate the result of the operation:
+
+|Status Codes|
+|---|
+|200|
+|204|
+|307|
+|400|
+|401|
+|403|
+|404|
+|415|
+|500|
+
+#### POST
+
+HTTP `POST` should be used to create new resources. It is a non-safe and non-idempotent operation and clients should be aware of that.
+
+Successful `POST` requests should always result in a `201` status code and empty response. The response should also contain a reference
+to the resource that was created:
+
+```
+HTTP/1.1 201 Created
+Content-Length: 0
+Location: http://mykeycloak.com/realms/{realms}/apis/users/{version}/1234
+```
+
+##### Expected Status Codes
+
+It is expected that APIs responding to HTTP `POST` requests respect the following HTTP status codes to communicate the result of the operation:
+
+|Status Codes|
+|---|
+|201|
+|400|
+|401|
+|403|
+|404|
+|409|
+|415|
+|500|
+
+#### PUT
+
+HTTP `PUT` should be used to replace a resource state.
+
+Successful `PUT` requests should always result in a `204` status code and empty response.
+
+```
+HTTP/1.1 204 No Content
+Content-Length: 0
+```
+
+#### PATCH
+
+HTTP `PATCH` should be used to partially update the resource state.
+
+Successful `PATCH` requests should always result in a `204` status code and empty response.
+
+```
+HTTP/1.1 204 No Content
+Content-Length: 0
+```
+
+Partial updates should follow the [RFC7396 - Json Merge Patch](https://tools.ietf.org/html/rfc7396) standard. 
+
+The `PATH` operation should be driven by a specific `Content-Type` header as defined by RFC7396 and other patch strategies. By using the
+`Content-Type` as a selector, this guideline can support additional patch strategies in the future.
+
+##### Note on RFC7396 - Json Merge Patch
+
+The RFC7396 is the most simple `PATH` strategy, but it has some limitations. The limitations are basically around updating lists on a resource which will be entirely replaced by the patch.
+
+Under certain circumstances, other patch strategies may be useful thus the constraint around the `Content-Type` header of `PATH` requests.
+
+##### Expected Status Codes
+
+It is expected that APIs responding to HTTP `PUT` requests respect the following HTTP status codes to communicate the result of the operation:
+
+|Status Codes|
+|---|
+|204|
+|400|
+|401|
+|403|
+|404|
+|409|
+|415|
+|500|
+
+#### DELETE
+
+HTTP `DELETE` should be used to remove a resource.
+
+Successful `DELETE` requests should always result in a `204` status code and empty response.
+
+```
+HTTP/1.1 204 No Content
+Content-Length: 0
+```
+
+##### Expected Status Codes
+
+It is expected that APIs responding to HTTP `DELETE` requests respect the following HTTP status codes to communicate the result of the operation:
+
+|Status Codes|
+|---|
+|204|
+|400|
+|401|
+|403|
+|404|
+|500|
+
+## Practices
+
+### Error Handling
+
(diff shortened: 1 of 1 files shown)


design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+However, it is important to taken into account the following principles when defining resource representations:
+
+* Keep it simple for clients, so that a representation can be easily built and processed when making requests/responses to the API
+* Be aware of extensive representations as they can affect overall performance and usability, if possible review your API to see if resources can be decomposed into smaller ones
+* Avoid deeply nested JSON objects
+* Avoid using generic resource representations
+
+### Http Status Codes
+
+The tables below defines the HTTP status code that clients should expect from APIs.
```
vmuzikar: In each table, should we also state what headers should be included in the response. It's not applicable for each of the stated codes but e.g. `405` response should contain `Allow` header, `307` code should have `Location` header, etc.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+
+|Verb|Description|
+|---|---|
+|400|Indicates that the request is invalid, usually related with the validation of the payload|
+|401|Indicates that clients should provide authorization or the provided authorization is invalid|
+|403|Indicates that the authorization provided by the client is not enough to access the resource|
+|404|Indicates that the requested resource does not exist|
+|405|Indicates that the method chosen by the client to access a resource is not supported|
+|409|Indicates that the resource the client is trying to create already exists or some conflict when processing the request|
+
```
vmuzikar: We should also cover code 415 Unsupported Media Type.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+|307|
+|400|
+|401|
+|403|
+|404|
+|500|
+
+##### Notes on `201` Status Code and Actions using `GET`
+
+`GET` can represent some action supported by a resource (see [Controller Resource](#controllerresource) for more details) and eventually
```
vmuzikar: The "Controller Resource" link is broken.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+create resources as a consequence of the action. In this case, developers can use `201` on `GET` responses to communicate to clients the result of the operation.
+
+Under this circumstance, the developer should consider the same principles for `POST` requests especially in regards to managing
+conflicts.
+
+##### Notes on `204` Status Code and Actions using `GET`
+
+`GET` can represent some action supported by a resource (see [Controller Resource](#controllerresource) for more details) and eventually start
```
vmuzikar: The "Controller Resource" link is broken.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+#### Server Side Error Codes
+
+|Verb|Description|
+|---|---|
+|500|Indicates that the server could not fulfill the request due to some unexpected error|
+
+### Http Verbs
+
+Based on the REST architectural style, HTTP Verbs should be used as the means to communicate to APIs the operations to be performed
+on their resources. While HTTP verbs cover by themselves the most common operations you might want to look the [Controller Resource](#controllerresource) as an alternative on how to overcome any limitation imposed by the usage of HTTP verbs herein defined.
```
vmuzikar: The "Controller Resource" link is broken.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+
+##### Notes on `201` Status Code and Actions using `GET`
+
+`GET` can represent some action supported by a resource (see [Controller Resource](#controllerresource) for more details) and eventually
+create resources as a consequence of the action. In this case, developers can use `201` on `GET` responses to communicate to clients the result of the operation.
+
+Under this circumstance, the developer should consider the same principles for `POST` requests especially in regards to managing
+conflicts.
+
+##### Notes on `204` Status Code and Actions using `GET`
```
vmuzikar: Should `GET` really lead to a resource changing action (even though it's through a controller)?

Let's take the GitHub's [`merge` controller](https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button) as an example. They're using `PUT` and it's only logical.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+
+`GET` can represent some action supported by a resource (see [Controller Resource](#controllerresource) for more details) and eventually start
+some processing asynchronously without necessarily return a response body. In this case, developers can use `204` on `GET` responses to communicate to clients the result of the operation.
+
+#### POST
+
+HTTP `POST` should be used to create new resources. It is a non-safe and non-idempotent operation and clients should be aware of that.
+
+Successful `POST` requests should always result in a `201` status code and empty response. The response should also contain a reference
```
vmuzikar: Should the response be really empty? IMHO the representation of the newly created resource should be returned.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+|403|
+|404|
+|409|
+|500|
+
+#### PUT
+
+HTTP `PUT` should be used to replace a resource state.
+
+Successful `PUT` requests should always result in a `204` status code and empty response.
```
vmuzikar: Should the response be really empty? IMHO the representation of the updated resource should be returned.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+
+```
+HTTP/1.1 204 No Content
+Content-Length: 0
+```
+
+##### Expected Status Codes
+
+It is expected that APIs responding to HTTP `POST` requests respect the following HTTP status codes to communicate the result of the operation:
```
vmuzikar: "... that APIs responding to HTTP ***`PUT`*** ..."

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+Successful `DELETE` requests should always result in a `204` status code and empty response.
+
+```
+HTTP/1.1 204 No Content
+Content-Length: 0
+```
+
+##### Expected Status Codes
+
+It is expected that APIs responding to HTTP `POST` requests respect the following HTTP status codes to communicate the result of the operation:
```
vmuzikar: "... that APIs responding to HTTP ***`DELETE`*** ..."

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+
+|Status Codes|
+|---|
+|204|
+|400|
+|401|
+|403|
+|404|
+|409|
```
vmuzikar: I don't think `DELETE` should respond with 409 Conflict.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+|500|
+
+## Practices
+
+### Error Handling
+
+A good error handling mechanism helps to communicate to clients the reasons why a request could not be fulfilled and eventually
+recover from this errors.
+
+When returning errors, developers should use a common [HTTP Status Code](#httpstatuscodes) and include a body using JSON as the format with the information
```
vmuzikar: The "HTTP Status Code" link is broken.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+When processing error responses, developers should look the error codes being used by other APIs and see if they can be reused.
+
+### Filtering
+
+TODO
+
+### Pagination
+
+TODO
+
```
vmuzikar: Should we also consider implementation of some sort of abuse rate ([similarly to GitHub](https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits))?

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+The goals of the guideline are:
+
+* Define the [overall principles](#principles) that should be taken into account when designing REST APIs for Keycloak
+
+* Define common practices and patterns for all Keycloak REST APIs such as:
+    * [Filtering](#filtering)
+    * [Pagination](#pagination)
+    * [Error Handling](#error_handling)
+    * [Documentation](#securityp)
```
vmuzikar: The following links are broken: Error Handling, Documentation

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+* Keep it simple for clients, so that a representation can be easily built and processed when making requests/responses to the API
+* Be aware of extensive representations as they can affect overall performance and usability, if possible review your API to see if resources can be decomposed into smaller ones
+* Avoid deeply nested JSON objects
+* Avoid using generic resource representations
+
+### Http Status Codes
+
+The tables below defines the HTTP status code that clients should expect from APIs.
+
+This document also defines which codes are expected to be used for each HTTP verb, please take a look at [HTTP Verbs](#httpverbs) for more details.
```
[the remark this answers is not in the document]
pedroigor: Fixed

design/rest-api-guideline.md
pedroigor: Fixed

design/rest-api-guideline.md
pedroigor: Fixed

design/rest-api-guideline.md
pedroigor: Fixed

design/rest-api-guideline.md
pedroigor: @abstractj, did some research about pagination. I did not have the chance to check his proposal at this regard. I'm open to discuss all the practices we find useful to base our own.

vmuzikar: This comment is not exactly about pagination. :) It's just about the abuse rate in the context of whole REST API. (I should've put the comment somewhere else so it wouldn't be confusing.)

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+Here is an example of a resource path related with the group of functionalities provided by the Account Management REST API:
+
+```
+/apis/accounts/v1/{user_id}
+```
+
+The path refers to the `accounts` group and any resource/operation related with user account management (in this case for user with id `user_id`) can be accessed from that path.
+Note that by grouping functionalities together and in conjunction with the versioning schema, we can keep the root path and also the semantics of the API without impact the adaptability for new changes or improvements to the API.
+
+Group names should be in lowercase, a single word name and plural.
```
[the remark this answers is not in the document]
pedroigor: As described, the group should represent a group of functionalities. These functionalities will probably operate on a top-level resource, even this top-level resource is logical.

The admin API example you mentioned is exactly what we are trying to improve. With the resource path format we are proposing, any Admin API we expose (e.g: users, groups, clients, identity providers, realms, etc) will have distinct groups and not include "admin" in any part of the resource path.

The path format herein proposed aims to create a uniform view of our APIs while keeping consistency on how they are exposed. Note that "/apis/admins" does not really provide any meaning for consumers, what they are interested in is how to manage resources. For instance, for users "/apis/users".

if you see the example section, you'll see how the User API should be exposed.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+* remove functionalities or specific parts of the API
+* provide non-stable or technology preview features
+* the different APIs can evolve separately
+
+#### Extensions
+
+To better acommodate and separate the core API from extensions built on top of the API, developers should consider the following path format for extensions:
+
+```
+/apis/{API_GROUP}/extensions/{version}
```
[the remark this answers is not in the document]
pedroigor: You usually write extensions to some existing functionality. For instance, extensions to {{account}} group should go under this group. The same goes for extensions related to authentication, authorization, identity providers, etc.

By using existensions on a per-group basis we are able to:
* Control lifecycle of extensions on a per-group basis
* Make clear which group the extension relates to

I do agree about making the format more clear by specifying the {{extension_id}} though.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+This separation aims to avoid confusion when consuming stable and official Keycloak APIs so that clients can differentiate if they are using some API provided by a third-party with no support from Keycloak team.
+
+#### Technology Preview
+
+To better acommodate and separate the core API from technology preview features, developers should use a beta version as the API version:
```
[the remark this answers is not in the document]
pedroigor: The proposed format also makes clear that the API is not supported. The work to change a root path in an application should quite trivial. Note that this is all about the root path. The sub-resources would still be the same when we move a beta API into our list of supported APIs.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+/apis/accounts/v1/1234/disable
+```
+
+The example above shows the usage of `disable` to operate accounts in order to disable them.
+
+When a resource fits under this category, the name should be a verb, using camel-case if necessary.
+
+### Resource Representation
+
+The standard format for resource representation is JSON. There is no rule for schemas or the semantics used by a resource representation, as long as they follow the JSON syntax properly.
```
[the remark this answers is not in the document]
pedroigor: Good point. To avoid confusion whether or not HATEOAS should be considered, I've included a `Level of Service Maturity` section that states we should be doing Level 2. Pretty much what we do today, no HATEOAS.

design/rest-api-guideline.md
pedroigor: Ahh sorry. I did associate it with the pagination practice.

Your suggestion seems to be more related with recommendations, what if we provide a third top-level concept to the guideline specific for that? We could probably use this part of the guideline to include any recommendation we find necessary for consuming our APIs.

abstractj: I believe rate limiting could be part of the Security section.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,388 @@
...
+The table below defines the HTTP verbs you should consider when exposing operations through your API and the circumstances where they apply.
+
+|Verb|Description|Idempotent|
+|---|---|---|
+|GET|Requests the current state of a resource|Yes|
+|POST|Creates a new resource or submit a command|No|
+|PUT|Replaces a resource state|Yes|
+|DELETE|Removes a resource|Yes|
+
+[TODO]: Shall we encourage PATCH for partial updates. That could make things easier for clients and avoid confusion when using PUT to update resources, where PUT
```
abstractj: +1 to include PATCH

design/rest-api-guideline.md
vmuzikar: >I believe rate limiting could be part of the Security section.
+1

design/rest-api-guideline.md
vmuzikar: +1

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: Ok, I think I get the idea now - the API "group" is basically a resource (in terms of REST). But then why to call it a group when it's a resource? Isn't it a bit confusing? :)

One thing I don't understand now is why the API version is present after the resource (`/apis/{API_GROUP}/{version}`)? Are we going to have a different versioning for each resource? Isn't it a best practice to have a single unified version for the whole API (to not overcomplicate things)?

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: Ok, I agree - makes sense to have extensions grouped under a resource/API group.

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: Thanks for adding that section. I believe the HATEOAS question is clear now.

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: Does it mean that once we change our support level for given version (i.e. we decide that some preview versions is now good enough to be supported), the old API just stop working?

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: Yeah, the group is a root resource which will probably have correlate subresources. So, a group of resources. If you have a better word for this, I'm happy to change it.

Regarding versioning, let me give you a concrete example. As you know we have a v2 of the account console which will become the official version very soon. Previously we had v1 which was not backed by a REST API, but if it were, the versioning scheme on a per-group/root resource basis would help. I think this model is quite flexible since we have different APIs that may evolve each one on its own pace. See, most of the time the version will be the same across all APIs, at least until some point in time, but when this is not true anymore we have the possibility to tell clients about breaking changes and new functionality being added for a specific API.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: Yeah, but just like any deprecation policy, we will give some time to clients to migrate.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: Thank you for all the valuable feedback.

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: I get your point but let me give you another example. :) Keycloak components. We have a unified version for all components in the server and (currently) even adapters as well. We don't have a different version for each SPI.
And I'm not sure it's very RESTful to put a version after (root) resource. (Personally, I haven't seen anyone else taking this approach. :))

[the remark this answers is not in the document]
pedroigor: Good example :) That is true and you will probably agree on about how much we are careful to change these SPIs.

The model I'm using is pretty much similar to how kubernetes or even google. For the latter, you may argue that each API is indeed a separated thing but that is the idea behind these groups.

If you want more details about how k8s look this https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/api-group.md.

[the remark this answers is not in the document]
stianst: I believe there will be multiple APIs (API group). I don't see these as a top-level resource, but rather a specific API. Admin is one, account is another, etc.. So I would argue it should not be plural (admin, account).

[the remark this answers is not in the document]
vmuzikar: @stianst +1
It then imho makes sense to have independent version for each API.

design/rest-api-guideline.md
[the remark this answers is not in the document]
stianst: I think this depends on whether or not the "api group" supports extensions or not, some may others may not. We would also need extensions at the top level so something like "/apis/extensions/{EXTENSION_ID}".

design/rest-api-guideline.md
```diff
@@ -0,0 +1,388 @@
...
+    - [Filtering](#filtering)
+    - [Pagination](#pagination)
+    - [Documentation](#documentation)
+  - [Examples](#examples)
+
+## Principles
+
+This section defines the general principles that should be considered by API developers, regardless of the practices defined through this document.
+
+### Level of Service Maturity
```
stianst: Please expand a bit on this section so those unfamiliar with "maturity level 2" doesn't have to refer to external description of it.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,388 @@
...
+
+Based on [Richardson Maturity Model](https://martinfowler.com/articles/richardsonMaturityModel.html), developers MUST adhere to the maturity Level 2, where interactions with the API endpoints are based on the usage of HTTP Verbs and status codes to communicate status and errors.
+
+### Resource Path
```
stianst: On one side I do like this, but there are at least two issues with it:

* admin endpoints are special and folks will often want to only expose these internally, both console and endpoints. With this there would be '/admin' for the console, with '/apis/admin/' for endpoints, which both have to be blocked for external traffic (or bound to internal IP/separate port).
* At some point I would like to be able to support virtual names for realms. For example allow myrealm.company.com to map to myrealm. That would require the realm to be the first part in the URL.

I suggest:

* '/apis/{API_GROUP}' - for APIs that do not operate on a specific realm (I don't think we have or will have any though, so probably not needed)
* '/realms/{REALM}/apis/{API_GROUP} - for APIs that operate on a specific realm (I believe this actually applies to all APIs)

design/rest-api-guideline.md
```diff
@@ -0,0 +1,388 @@
...
+
+```
+/apis/accounts/v1/1234/applications
+```
+
+The example above is representing a resource store path from where we can manage the `applications` for a user account with ID `1234`. Based on this path, clients can perform different operations based on the HTTP verbs or any controller resource defined to the store. Usually this type of resource is associated with CRUD operations for a specific resource.
+
+When a resource fits under this category, the name should be a plural noun.
+
+#### Controller Resources
```
stianst: Should we not be more prescriptive with regards to controller resources? I'd like to see one way of invoking an operation, not multiple. Either GET or POST, not a choice, etc.

design/rest-api-guideline.md
stianst: PATCH is a must! I would also like to see this guideline cover how we will support patch for JSON (perhaps looking at existing RFCs like https://tools.ietf.org/html/rfc7396, https://tools.ietf.org/html/rfc6901, https://tools.ietf.org/html/rfc6902).

Partial updates in the current admin endpoints is one of the messiest and inconsistent parts of the API. Sometimes it does partial updates, sometimes it doesn't. Sometimes it'll add to an array, sometimes it'll replace the array, etc..

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: Added to the TODO list.

design/rest-api-guideline.md
pedroigor: PATH is already in the TODO list.

I agree with you and that is the reason why I've left this open for discussion. PATH will help to communicate better and consistently how we perform updates together with `PUT`.

However, I don't think we should strictly follow those standards you mentioned, but something simpler where the payload sent using PATH does only partial updates. We are kind of doing RFC7396 already in some places, so we should stick with it. The other specs will make things more complicated, IMO. For both consumers and API developers.

design/rest-api-guideline.md
pedroigor: Well, that goes against the idea of the guideline and similar to current resource path format. I was expecting this discussion as this is one of the most critical points of the guideline.

Firstly, I don't think that `/apis/{API_GROUP}' is not needed. If you look at the example section, you'll see how we would use the format defined by this guideline with the User API. Let me put it here:

|Resource Path|GET|POST|PUT|DELETE|
|---|---|---|---|---|
|/apis/users/v1/realms/acme|List all users in the realm `acme`|Creates an user in the realm `acme`|Not Supported|Not Supported|
|/apis/users/v1/123|Gets a representation for user with id `123`|Not Supported|Updates the user with id `123`|Deletes the user with id `123`|

I really think that this format is much more descriptive, concise and aligned with our functionalities. It provides a great separation and  a versioning schema on a per-api basis. More justifications you'll find in the `Group` section of the document.

Also, note that you don't actually need the realm in the path when managing a specific user. You can perform all operations for a particular user just based on its ID.

The same goes with `/apis/accounts`, why you need the realm if you are operating in the context of a specific user ? The realm is implicit.

Regarding the protection of the APIs part of the admin logical group. I don't think the path format impacts so much on how you would protect these APIs. You could, for instance, blacklist APIs and whitelist only those you want to expose.

Regarding the virtual names. You kind of push me to the corner with this one :) I wasn't aware of this requirement and is it so special to force us to use a specific path format and ignore other options (eventually better)?

Taking the path we use for login as an example, today we have `/realms/{realm}/protocol/openid-connect/auth`. With the format I'm proposing the path would be `/apis/auth/v1/{realm}/openid-connect` (ok, we don't have a plural in the api group name here :)). Where `auth` is the API group and the version schema is specific for the Authentication API which can change in order to support, for instance, future and breaking versions of openid-connect.

design/rest-api-guideline.md
pedroigor: I will try to add more info. I tried to give an example using a fictional `/apis/accounts/v1/1234/disable` operation.

I think whether to use GET or POST, depends on the operation. If that is the type of thing you want to have more clarity in this section.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: Based on the idea of the guideline, API groups can be seen as a top-level resource. Thus the reason for having a plural name.

There are some examples in guideline for the account case.

Regardgim Admin, that is something I'm trying to avoid, and I'm happy to make this more clear. All sub-resources that we have today for things like users, etc, will have their own api group instead of subresources of an API group such as 'realms' or 'admin'. Instead of `/realms/{realm}/users` we would have just `/apis/users`.

The "admin" is a logical group and not really part of the API path.

pedroigor:
Added more items to the TODO list.

design/rest-api-guideline.md
stianst: There is no clear contract in how we do partial updates today. We should not invent our own JSON patch approach, but rather stick to something already defined. There's even Java implementation of this stuff that allows you to do it fairly easily. What we have today is a confusing mess that is beyond inconsistent, so the alternative to using an existing spec is to invent our own, which is a bad idea.

design/rest-api-guideline.md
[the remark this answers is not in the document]
stianst: The name should be v1preview1 or something, not beta1. There is a fundamental difference between the two when it comes to support. As long as the API is in preview we should not provide any deprecation policies, but simply remove and replace. This should be made clear to people though.

design/rest-api-guideline.md
stianst: I was not expecting a URL scheme to be defined as part of the REST guidelines. That is an additional bonus, but it does introduce a lot more to discuss ;)

I see absolutely no benefits to putting the realm in the place that you have. It's confusing and wrong. As a user I do not know when to include it and when to not include it. I can't simply prefix URLs including the realm as is possible today. I can't add support for virtual host for realms (sorry, but this is something I want to at least be able to add in the future).

Also, we always need to know what realm we are operating under. A request to /apis/users/v1/123 needs to know the realm in order to know how to verify the bearer token. Further, sure today a user id is unique as it is a uuid. That is not a given that it will be in the future. We may even have users loaded from different locations for different realms.

As a realm is our tenant it should always be specified and should be a top-level thing.

We can continue the discussion around this if you want, but you're going to have a very hard time to convince me here.

design/rest-api-guideline.md
stianst: I want it as prescriptive as possible, with as little flexibility as possible. I'm fairly certain that we can limit it to POST only. We had some GET based operations in the past, which all as far as I know, have now been converted to POSTs due to bugs, restrictions and security issues (exposing sensitive stuff as query params).

design/rest-api-guideline.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+#### Error Codes
+
+Error codes can be any string. They should be lowercase and use `_` when the code has multiple words. However, developers should 
+consider existing error codes before creating new ones.
+
+When processing error responses, developers should look the error codes being used by other APIs and see if they can be reused.
+
+### Filtering
```
stianst: Is Filtering the same as search? I.e. if I want to find a user by username, or list users with a given role?

design/rest-api-guideline.md
```diff
@@ -0,0 +1,388 @@
...
+    * [Documentation](#documentation)
+
+* Define common practices for [documentation](#documentation)
+
+Although the guideline is based on the REST architectural style, it does not try to strictly adhere to RESTful constraints. The principles covered by the guideline are based on practical experiences in the development of REST APIs in Keycloak while still conforming with most of the constraints of REST architectures.
+
+**Table of Contents**
+
+
+  - [Principles](#principles)
```
stianst: We should also consider a way to limit what fields are returned, which is especially relevant for larger objects. In the future I can see us having a single realm json config that contains all config for the realm, instead of separate child resources for user storage providers, etc. Then rely on PATCH to add/remove to the realm config.

design/rest-api-guideline.md
pedroigor: OK, you have a decision then and I won't continue the discussion. I have a different opinion, I don't think is wrong (can give you well-known examples), and I do see benefits with this approach.

I'll update the documentation accordingly.

design/rest-api-guideline.md
pedroigor: Yes, it is.

design/rest-api-guideline.md
pedroigor: Sure, I'm not saying that there is a clear contract today. But that RFC7393 is what I think we should consider for partial updates. Which we do in some way in some of our APIs.

design/rest-api-guideline.md
pedroigor: Usage of `GET` should consider any security recommendation. But if the actions we are envision for now are all based on `POST`, we can just omit `GET` for now.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: True. So I guess we can avoid versioning in this case and just use `/preview/{api}` to communicate a tech preview API.

design/rest-api-guideline.md
pedroigor: The filtering part should be covered by the filtering section so that you can fetch only the fields you want (user storage providers) and then use PATCH to update only those fields.

design/rest-api-guideline.md
pedroigor: I will leave that open of others comment and see how far we want to go. IMO, some of these codes have implicit headers.

design/rest-api-guideline.md
pedroigor: Probably not. Will change this. I'll remove GET as a supported action method and recommend POST or PUT instead.

design/rest-api-guideline.md
pedroigor: The guideline states that a Location header should be returned pointing to the newly created resource. Considering that the location path includes the ID, clients can either use it to reach the resource that was created or extract the id from the location path in order to update the representation they already used to create the resource.

The point here is that we are avoiding unnecessary responses and saving network/processing.

design/rest-api-guideline.md
pedroigor: When you do a put, you already have a representation of the resource being updated, so why do you think we should return it back ?

Again, the point here is to save some bandwidth and processing.

design/rest-api-guideline.md
pedroigor: Based on @vmuzikar comments, we'll use either `POST` or `PUT` for actions. Maybe `PUT` could be used for cases where the action creates additional resource or just updates the resource ? So the client don't need to think about which one to use ?

pedroigor:
@vmuzikar, I don't know why but your latest comments could only be seen by iterating over the file change set. That is why I took so much time to give you a response.

design/rest-api-guideline.md
stianst: I would just go with POST for all actions. In a way you can read it as "post a new action to be executed" ;)

pedroigor: I don't have any preference. I can read PUT in a similar way. I was just wondering whether or not PUT is more correct given that we are operating, and eventually changing, an existing resource.

I can live with POST and just remove references to GET from the document.

pedroigor:
@stianst @vmuzikar @abstractj, please check the TODO list for updates to the document.

pedroigor:
@stianst, still regarding the resource path format. What do you think about this, taking the User API as an example: `/realms/{realm}/apis/users/v1` and `/realms/{realm}/apis/users/v1`. Note that there is no reference to `admin`. Like I mentioned before, `admin` could be just a logical concept and the protection of admin related APIs could still be easily achieved.

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: @pedroigor That means the "preview" part of the path would be in place of a version? So instead of `/v1/users/{id}` would be something like `/preview/users/{id}`?
If so, +1 :)

stianst:
> @stianst, still regarding the resource path format. What do you think about this, taking the User API as an example: `/realms/{realm}/apis/users/v1` and `/realms/{realm}/apis/users/v1`. Note that there is no reference to `admin`. Like I mentioned before, `admin` could be just a logical concept and the protection of admin related APIs could still be easily achieved.

I'm not too keen on dropping admin. Users is one thing, but are you assuming the same for everything else? Seems we'll have a lot of "API groups" that are related to admin operations, but are not obvious that they are related. Also, how would you imagine someone sets up firewall restrictions on these things?

design/rest-api-guideline.md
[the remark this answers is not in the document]
stianst: +1

design/rest-api-guideline.md
vmuzikar: Maybe the question should be: "how often do you want to fetch a newly created resource?"
I'd say it's rather common that you create something (with some missing fields, like ID) and then you want to see the result with all the (generated) fields. And in this case we wouldn't be saving network - quite the opposite.

design/rest-api-guideline.md
vmuzikar: Yeah, you're right it doesn't make much sense to return the resource when doing `PUT`.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,406 @@
...
+```
+HTTP/1.1 204 No Content
+Content-Length: 0
+```
+
+#### PATCH
+
+HTTP `PATCH` should be used to partially update the resource state.
+
+Successful `PATCH` requests should always result in a `204` status code and empty response.
```
vmuzikar: Again, should the response be empty? It IMHO makes sense to return the final result when doing a partial update.

design/rest-api-guideline.md
vmuzikar: IMHO `PUT` would be more appropriate for triggering an operation using a controller.

The operation is (usually) a "shortcut" for a resource modification. And `PUT` is intended to be used for modifications.

We can also look at this from a different perspective. Are controller operations idempotent? I believe they should be and therefore `PUT` is the obvious choice (`POST` is not considered to be idempotent).

design/rest-api-guideline.md
vmuzikar: We may also consider taking similar approach to GitHub: by default, return only some limited set of fields (basic info); if you need more, you have to specify those extra fields in the filter.

vmuzikar:
@pedroigor If I may suggest something, could you not force push to this PR and leave the commits separated until it's ready to be merged? :) This would make it a lot easier to track the changes (I mean, this is still WIP/draft, so the commits don't need to be squashed yet).

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: Yes. Like this `/realms/{realm}/apis/{API_GROUP}/{version}/preview/{preview_id}` or `/realms/{realm}/apis/preview/{preview_id}`.

design/rest-api-guideline.md
pedroigor: +1

design/rest-api-guideline.md
pedroigor: Not always and that is why I mentioned that for me PUT looks more correct. You know what, I think we should leave this open so you could either use PUT or POST. People should respect idempotency (which is already covered in the guideline) when choosing a method.

I was thinking a bit more about GET and despite the security considerations inherent to this method, I think it should also taken into account. For instance, I can think about a GET action that starts some process through redirection.

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: @pedroigor Actually, what I meant was that `preview` is the version identification. So no `{preview_id}` should be required. Because, if I understand it correctly, there will be at most only one preview version available at the same time. And there will be no deprecation policy - when a feature is not a preview anymore, it's just gone from the `preview` version.

In other words, instead of `v1`, `v2` and similar, you would use `preview` (instead of the actual version).

design/rest-api-guideline.md
pedroigor: In my previous reply, I mentioned how you could obtain the ID from the response. See the POST response in the guideline, isn't that enough for most cases? In our testsuite, for instance, there are places where this is enough.

When you create a resource you have the representation. If you want the ID is just a matter of checking the `Location` header and extract the ID. Or make another request if you really want to fetch the resource state.

design/rest-api-guideline.md
pedroigor: I'm not sure, because it is a very specific operation. I'll leave open so others can give their opinion.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: The `preview_id` is a name of the preview future. In your example, it would be `users`. This is similar to the extension format.

pedroigor:
> I'm not too keen on dropping admin. Users is one thing, but are you assuming the same for everything else? Seems we'll have a lot of "API groups" that are related to admin operations, but are not obvious that they are related. Also, how would you imagine someone sets up firewall restrictions on these things? Today it is as simple as blocking /admin/ for external IPs, what does that look like with the proposed format?

You could use a hybrid approach in your WAF, blacklisting everything by default and whitelist only the endpoints you want to make public. I guess we have quite a few endpoints that should be made public, right ? Such as `/auth`, `/account`, etc.

But yeah, I won't push for this. I just wanted to see your thoughts about the format I mentioned so we could remove one more thing from the resource paths.

pedroigor:
> @pedroigor If I may suggest something, could you not force push to this PR and leave the commits separated until it's ready to be merged? :) This would make it a lot easier to track the changes (I mean, this is still WIP/draft, so the commits don't need to be squashed yet).

I'll ... Force of the habit :)

design/rest-api-guideline.md
```diff
@@ -0,0 +1,434 @@
...
+consider existing error codes before creating new ones.
+
+When processing error responses, developers should look the error codes being used by other APIs and see if they can be reused.
+
+### Filtering
+
+TODO
+
+### Pagination
+
```
abstractj: @pedroigor I don't know what you think about the proposal below. But let me know if makes sense:

### Pagination

For large sets of data, since the early releases pagination and filtering capabilities were [provided on Keycloak](https://www.keycloak.org/docs-api/6.0/rest-api/index.html) by using a cursor (*first*) and page size (*max*).

For example, to retrieve a list of the first 10 groups starting at the first item we provide:

* Request

```
curl -I "http://localhost:8081/auth/admin/realms/master/groups?first=0&max=10" -H "Authorization: bearer $TKN" -H "Accept: application/json" -H "Authorization: Bearer $TKN"
```
* Response headers

```
HTTP/1.1 200 OK
Connection: keep-alive
Cache-Control: no-cache
Content-Type: application/json
Content-Length: 961
Date: Wed, 22 May 2019 13:44:57 GMT
```

#### Possible enhancements

Today, the response headers do not provide too much detail about the number of items we have into the database. Adding that would be handy for the new account console and also other people [integrating with our APIs](http://lists.jboss.org/pipermail/keycloak-user/2019-May/018108.html).

Please, notice that this is just an enhancement to the current implementation approach as we have the opportunity to design something new. For example:

* Request

```
curl -I "http://localhost:8081/auth/admin/realms/master/groups" -H "Authorization: bearer $TKN" -H "Accept: application/json" -H "Authorization: Bearer $TKN"
```
* Response headers

```
HTTP/1.1 200 OK
Connection: keep-alive
Cache-Control: no-cache
Content-Type: application/json
Content-Length: 961
Date: Wed, 22 May 2019 13:44:57 GMT

Link: <http://localhost:8081/auth/admin/realms/master/groups&page=2>; rel="next",
  <http://localhost:8081/auth/admin/realms/master/groups&page=40>; rel="last"
```

[GitHub uses this approach](https://developer.github.com/v3/guides/traversing-with-pagination/) through [RFC 5988 for Web Linking](https://tools.ietf.org/html/rfc5988), which is mentioned in the HATEOAS section from the [REST API Tutorial](https://restfulapi.net/hateoas/). Web Linking is something that already comes with JAX-RS API.

Alternatively, people should be able to query the number of items on a page.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,434 @@
+# REST API Guideline
+
+* **Status**: Draft
+* **Author**: keycloak-team
```
abstractj: @pedroigor I'd suggest to add you as author (if you would like to) and add an extra field with "Co-authors" by putting people's name on it. I'd rather to know who to contact, I believe keycloak-team is too generic.

justintime4tea:
Another thing that just came to mind as well is rate limiting. I would love to see some form of the leaky bucket or other well battle tested pattern and methodology around limiting API calls either on a per client or per authenticated user basis with a sane default. Having rate limiting baked in would be awesome IMO.

pedroigor:
> With regards to resource manipulation no mention of concurrency handling is made. Some RESTful APIs implement optimistic concurrency where a resource is given a version and that version must be included with subsequent state changing operations. When the version accompanying the REST request differs from what is "current" we can assume it has been changed elsewhere. This typically results in a 412 status of precondition not met version mismatch. This gives the client an opportunity to present a user (if there is one) a message indiciating the resource was updated after they GET it but before they PUT/PATCH it and they should refresh or display the changes and confirm the modification. You see this with JIRA and other multi-user web tools a lot.
> 
> Other forms of concurrency handling like pessimistic (locking) is unfavorable (subjectively) as it adds much more complexity (speculative) in both server and client (managing lock).
> 
> Can you propose a "suggested" practice / pattern for handling concurrency within a resource? I've had great experience with optimistic concurrency but would love to hear your thoughts.

That is a good point, @justintime4tea. I will add an empty section for this as a TODO.

Please, feel free to contribute to this document with your experiences. From my side, I would expect that we achieve this by:

* As you suggested, optimistic concurrency is the best approach
* Use a version for each resource returned by the server which is completely opaque to clients and that should be sent back to the server when performing operations that modify the resource state
* I would like to know what you are using for controlling modifications to the state. I can think of using a simple index that is incremented and saved for each operation that modifies the resource state.
* How we communicate the version between clients and server? What is your experience with that? For me, it makes sense to the version to be part of the resource representation and avoid using headers (e.g.: If-Match or If-Unmodified-Since) or HTTP parameters. I think is less work for clients as they just need to deal with the representation as returned by the server.
* I still see value in supporting the version being passed as a HTTP parameter, but that is only for situations where (and if we going to provide that) clients can poll resources from the server. For instance, watch for modifications, etc.

Please, let me know what are your thoughts.

pedroigor:
> Another thing that just came to mind as well is rate limiting. I would love to see some form of the leaky bucket or exponential back-off or other well battle tested pattern and methodology around limiting API calls either on a per client, per host, per authenticated user or any combination with a sane default.
> 
> Having rate limiting baked in would be awesome IMO not just for potential abusers but for unexpected problems in API callers who even accidentally DDoS an API.

Another topic we should add here. In fact, we have a discussion started on the OAuth2 Device Grant design proposal. One of the suggestions made by @stianst was per client approach.

The issue with rate limiting is that you need different barriers as they alone are not enough. All have cons when used alone.

I don't have a strong opinion on that but if we want something that really helps we should probably introduce all the barriers you mentioned with a good configuration on how they are enforced.

design/rest-api-guideline.md
pedroigor: @abstractj, I changed because this is teamwork even though I've started the document/discussions. In any case, we have our lists which people can use to contact us.

But if you want me to change the author, I can do it too.

design/rest-api-guideline.md
pedroigor: The approach looks correct and aligned with some common practices we see out there.

I'm not sure if we really need to calculate the total count of entries and what exactly is the last page. Today our pagination have options for `First`, `Previous` and `Next Page`. We basically use the {{first}} and {{max}} parameters which I think is enough.

However, there is a behavior that we could certainly improve. Today, if you try to click the `Next Page` button you may get a empty result set. I would suggest that we perform queries based on N + 1 where N is the number of items on a page and use something to indicate the client that there is no next page. Using a {{Link}} header as you suggested may do the trick.

stianst:
I would also like to get validation covered as part of this. Specifically how validation results are returned. We will most likely want to have support for corse and fine validation message. For the latter a list of properties that didn't pass validation and corresponding validation error messages for each.

pedroigor:
I can add more info about that on the `Error Handling` section, but what we have here is basically an error message for each validation that did not pass thus an error on a per validation basis.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+The tables below defines the HTTP status code that clients should expect from APIs.
+
+This document also defines which codes are expected to be used for each HTTP verb, please take a look at [HTTP Verbs](#httpverbs) for more details.
+
+Clients should consider the status codes herein defined in order to properly handle responses from APIs, recover from errors, and 
+understand the reasoning behind the API responses.
+
+#### Success Codes
+
+|Verb|Description|
```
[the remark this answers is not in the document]
abstractj: @vmuzikar marking is as resolved because Pedro already updated as suggested.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+|---|---|
+|200|Indicates the request completed successfully regardless the size of the response body|
+|201|Indicates the request completed successfully and a resource was created|
+|204|Indicates the request completed successfully and the server did not provide a response body|
+
+#### Redirection Codes
+
+|Verb|Description|
```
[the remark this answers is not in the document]
abstractj: @vmuzikar marking it as resolved because it was already updated as suggested.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,389 @@
...
+|---|---|
+|307|Indicates that clients should resubmit the request to another location|
+
+#### Client Side Error Codes
+
+|Verb|Description|
```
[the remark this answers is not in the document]
abstractj: @vmuzikar marking it as resolved, because it was already updated as suggested.

design/rest-api-guideline.md
abstractj: @vmuzikar marking it as resolved. I tested the link myself and it is working now. Thanks for reporting it.

design/rest-api-guideline.md
abstractj: @vmuzikar marking it as resolved because this section no longer exists.

design/rest-api-guideline.md
abstractj: @vmuzikar marking as resolved, because it was already updated. Thank you.

design/rest-api-guideline.md
abstractj: Marking as resolved, because it was already updated.

design/rest-api-guideline.md
abstractj: @vmuzikar I totally understand your point, the GitHub API follow a similar approach. But I would like us to try what Pedro suggested. This is recommend as best practice in https://restfulapi.net/http-status-201-created/.

Also, let's keep in mind that nothing is set in stone here. This draft will be reviewed by keycloak-dev mailing list and others. So if we reach the conclusion that's not gonna work, we can always revisit.

design/rest-api-guideline.md
abstractj: @vmuzikar marking as resolved as it seems there's an agreement on it.

design/rest-api-guideline.md
abstractj: @vmuzikar I think it's too much. As Pedro mentioned, some of these codes are implicit. Let's take as an example HTTP 405. If we look at RFC 7231 (https://tools.ietf.org/html/rfc7231#section-6.5.5) we will find out that `Allow` will be present in the header.

vmuzikar: @abstractj Ok, I agree - let's not overcomplicate this. You can mark this as resolved (I don't seem to have the permissions).

design/rest-api-guideline.md
abstractj: > The approach looks correct and aligned with some common practices we see out there.
> 
> I'm not sure if we really need to calculate the total count of entries and what exactly is the last page. Today our pagination have options for `First`, `Previous` and `Next Page`. We basically use the {{first}} and {{max}} parameters which I think is enough.

@pedroigor I just updated the idea. But in summary we're not going to calculate the total of entries. But let consumers know which one is the last page, so they have an idea about how many items are available to consume. This can be useful for people, because they don't need to iterate over all the entries to know the amount of data they gonna consume. This is what happens today, people iterate over it until they hit an empty response.

Alternatively people could provide also provide `max` as we already do today. I can update my comments if you think it makes sense.

> 
> However, there is a behavior that we could certainly improve. Today, if you try to click the `Next Page` button you may get a empty result set. I would suggest that we perform queries based on N + 1 where N is the number of items on a page and use something to indicate the client that there is no next page. Using a `Link` header as you suggested may do the trick.

I'm not sure if I follow your idea. But let me try to elaborate how it was supposed to work, maybe we are talking about the same things. `rel="next"` will always return N + 1, so let's assume a scenario where the last page is 50 and I'm on page 49. That means that when I hit the page 49 I should get:

Link: <http://localhost:8081/auth/admin/realms/master/groups&page=50>; rel="next",
  <http://localhost:8081/auth/admin/realms/master/groups&page=50>; rel="last"

So if we hit next, the interface should be aware that page 50 is the last based on the response headers.

Does make any sense to you?

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: @pedroigor I'm still not sure I fully understand what `preview_id` should be. It seems to me a bit confusing, especially when we use `version` at the same time. Could you give me some specific real-life example how the full API path would look like, what `preview_id` could actually be?

Aren't previews often used in place of a version, something like `/{realm}/apis/accounts/v1preview1/{user_id}`?

design/rest-api-guideline.md
```diff
@@ -0,0 +1,444 @@
...
+
+#### Technology Preview
+
+To better accommodate and separate the core API from technology preview features, developers should use a path format as follows:
+
+```
+/{realm}/apis/{API_GROUP}/{version}/preview/{preview_id}
+```
+
+Technology preview APIs are related with features that have great potential to become fully supported. They provide the latest work in some area of interest and are offered with the intent to gather feedback from the community about new functionalities. However, such functionalities can be remove at any moment and break existing clients consuming the API.
```
vmuzikar: Typo, missing "d" in "remove":
>such functionalities can be remove**d**

design/rest-api-guideline.md
vmuzikar: @abstractj @pedroigor Ok, fair enough. :) Let's mark this as resolved.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: There are two scenarios here.

Consider you are adding a new feature set that you want to bake in the community. If the feature set is a new API Group you will have a path like:

```
/{realm}/apis/preview/newfeatureset
```

Now, if the feature set is added to an existing API Group such as `accounts`, you will have something like:

```
/{realm}/apis/accounts/v1/preview/newfeatureset
```

Is it more clear now? See that we decided to only version API Groups (e.g.: accounts).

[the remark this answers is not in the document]
vmuzikar: @pedroigor Thanks for clarification! So, if I understand it correctly, we won't use the preview "labeling" when modifying existing feature sets? It'll be applicable only for new feature sets that didn't exist before?

Say, we want to remove some field from user accounts resource. Would we somehow mark such change as a preview?

design/rest-api-guideline.md
pedroigor: In order to return `rel=last` you need to calculate the last page somehow. I don't understand how you are going to do that given that you also said we are not going to calculate the total of entries.

My last comment is about returning *only* `rel=next` to indicate to the client whether or not there is a next page. If the response contains the header the client can then render the `Next` button given that the server is indicating that there is more pages. Otherwise, the client can disable/hide the button as it reached the last page.

Considering your scenario where the client asks for page 49. Suppose `max` is 20. The server will query for 21 entries (starting on `first` which represents the actual page). If the query returns 21 it means we have a `rel=next`. Otherwise, there are no more pages and you don't return `rel=nex`.

design/rest-api-guideline.md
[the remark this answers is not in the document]
pedroigor: In this case, if the change can break backward compatibility, I think best is have a new version of `accounts` API Group. It does not fit into preview as we are actually modifying the API and not including a tech preview feature.

design/rest-api-guideline.md
abstractj: > In order to return `rel=last` you need to calculate the last page somehow. I don't understand how you are going to do that given that you also said we are not going to calculate the total of entries.

I got your point now, I thought your comment was related with my last sentence where the idea of returning the total. And of course there are performance penalties when we calculate the total of records just to display the last page.

For now, let's remove it and if more people request we can always iterate over it.

I just updated it, let me know if that works for you.

> 
> My last comment is about returning _only_ `rel=next` to indicate to the client whether or not there is a next page. If the response contains the header the client can then render the `Next` button given that the server is indicating that there is more pages. Otherwise, the client can disable/hide the button as it reached the last page.
> 
> Considering your scenario where the client asks for page 49. Suppose `max` is 20. The server will query for 21 entries (starting on `first` which represents the actual page). If the query returns 21 it means we have a `rel=next`. Otherwise, there are no more pages and you don't return `rel=nex`.

pedroigor: @abstractj, updated the PR with what we discussed.

design/rest-api-guideline.md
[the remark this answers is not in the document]
vmuzikar: @pedroigor Ah, I think I understand now! Thank you for your patience. :)
Technical Preview is actually not related to versioning. It's a tech preview in the same sense as we have in tech previews in Keycloak - it's just a separated feature set. Am I correct this time? If so, consider this resolved.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,476 @@
...
+    The position of the first result to retrieve
+
+* `max`
+    
+    The maximum number of entries to retrieve
+
+These parameters should be sent along with a `GET` request as follows:
+
+```
+/{realm}/apis/admin/users/v1?firt=1&max=20
```
abstractj: @pedroigor there's a minor typo here:  `firt`

design/rest-api-guideline.md
abstractj: @pedroigor thank you. Marking it as resolved.

design/rest-api-guideline.md
abstractj: @pedroigor I'm not picky about it and it's up to you. But I'd rather to just follow the RFCs approach, where we know the authors.

design/rest-api-guideline.md
abstractj: @stianst @pedroigor I would like to include the following content for filtering if possible:

### Filtering

Sometimes, we need more control over the content that will return from the server, so that only a subset of the data is available in the response. For example, a list user's groups or permissions. 

The Keycloak admin REST API allows to search for users based on `e-mail`, `first name`, `lastname` and other optional parameters. However, such implementation is limited to a particular context (e.g, collection of users).

Filtering can be implemented as a query parameter, like we already did for the current REST API, with some enhancements. To filter on a field, should be as simple as the following request:

```
/{realm}/apis/admin/users/v1?group=users1&realm=master&sort=desc
```

Multiple filters should result in an implicit `AND`, so in our example `/v1?group=users1&role=role1&realm=master&sort=desc` would provide results based on `groups`, `roles` and `realm`. Now let's say we want filter based on more than one groups:

```
/{realm}/apis/admin/users/v1?group=in:users1,users2&realm=master&sort=desc
```

A brief explanation about the parameters:

|Name|Type|Description|
|---|---|---|
|sort|string|Sort the results on our query based on: date, ascending order and descending order|
|in|string|Query for values inside a collection. For example: users belonging to group1 `and` group2.
|or|string|Query for values inside a collection. For example: users belonging to group1 `or` group2.

---
### Minor consideration

* It is possible to provide more generic queries like GitHub does in the future, by using something like `/commits?q=repo:octocat`. But for now, I'd vote to not over complicate things and stick with simple. At the moment we don't have anything that justifies generic queries in the Account REST API

design/rest-api-guideline.md
```diff
@@ -0,0 +1,476 @@
...
+
+The goals of the guideline are:
+
+* Define the [overall principles](#principles) that should be taken into account when designing REST APIs for Keycloak
+
+* Define common practices and patterns for all Keycloak REST APIs such as:
+    * [Filtering](#filtering)
+    * [Pagination](#pagination)
+    * [Error Handling](#error-handling)
+    * [Documentation](#documentation)
```
abstractj: @pedroigor I would like to have this section removed and follow up on this in a different PR. Please, see: https://issues.jboss.org/browse/KEYCLOAK-10787

pedroigor:
I agree. Considering that we have the practices defined and updated based on comments in this PR, the missing parts are only some of the practices. We can handle the missing practices as separated PRs, IMO.

design/rest-api-guideline.md
```diff
@@ -0,0 +1,476 @@
...
+    - [HTTP Verbs](#http-verbs)
+      - [GET](#get)
+      - [POST](#post)
+      - [PUT](#put)
+      - [DELETE](#delete)
+  - [Practices](#practices)
+    - [Error Handling](#error-handling)
+    - [Filtering](#filtering)
+    - [Pagination](#pagination)
+    - [Concurrency Control and Consistency](#concurrency-control-and-concurrency)
```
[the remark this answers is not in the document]
pedroigor: I would prefer to not remove it because:

* Document is marked as `Status: Draft`. I don't see an issue if we still have TODOs in a draft.
* Having them in the guideline helps to understand much better the document structure and what we want to cover in the document

design/rest-api-guideline.md
pedroigor: I would prefer to not remove it because:

* Document is marked as `Status: Draft`. I don't see an issue if we still have TODOs in a draft.
* Having them in the guideline helps to understand much better the document structure and what we want to cover in the document

design/rest-api-guideline.md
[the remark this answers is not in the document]
abstractj: No problem, let's leave it as is.

abstractj:
@stianst @vmuzikar @justintime4tea thank you for your the review and bringing valuable considerations to this Draft. Because it's a draft, and when it comes to REST best practices there's a lot to explore. We aim to address these considerations in individual PRs.

For now I'm merging it.

@pedroigor thanks for writing this first draft.

PR #8, opened 2019-05-23
Operator proposal

slaskawi:


design/operator.md
```diff
@@ -0,0 +1,64 @@
...
+providing seamless experience for OpenShift 4. An Operator simplifies deployment and
+management of all the services targeted for OpenShift platform.
+
+An Operator follows the Reconsiler Pattern (described in the Cloud Native Infrastructure book).
+It creates a target infrastructure state, gathers information about the current state and then reconsiles
+based on it.
+
+For more information, please take a look at the [Operator SDK documentation](https://github.com/operator-framework/operator-sdk).
+
+## Operator capabilities
```
iankko: @slaskawi  Would it be possible to add a feature, the operator to handle image upgrade from previous image version to the latest one? Suppose request to upgrade from RH-SSO 7.2.* (e.g. 7.2.4) image to the latest RH-SSO 7.3 one.

Maybe this is something that could be handled in pretty straightforward / universal way by the operator too. The idea / expectation here being, the operator (or subpart of it) would be able to check the current RH-SSO image version running, find out if there's an update available. If so, suggest the upgrade, and if confirmed / approved, it would update both, the image and the corresponding template to the latest available version.

WDYT?

P.S.: Assuming the request to update the running pods to the latest version would be a slightly more complex task, but it could update the image & the template, so the new apps / pods would be already started utilizing these.

design/operator.md
```diff
@@ -0,0 +1,64 @@
...
+
+Here’s a short list and a brief description what Operator can do today:
+
+### Keycloak provisioning
+
+The Operator can provision (and deprovision) RHSSO instances based on Keycloak CRD. The Operator will also spin up a Postgresql database and will create a Secret with an Admin credentials. It is also possible to attach different RHSSO plugins (such as metrics SPI plugin) and finalyzers (that are responsible for cleaning up the resources (routes, PVs etc) allocated during provisioning phase).
```
iankko: `s/finalyzers/finalizers/g` ?

design/operator.md
stianst: Seamless upgrades should be covered later. We are targeting the basic install capabilities in this round as the most important thing right now is to get started with something.

slaskawi: I added this to the `Further work` paragraph. Thanks for suggestion.

design/operator.md
slaskawi: Fixed thanks!

slaskawi:
> Explicitly mention that we are targeting basic install capabilities only in the first round, with a note in future work section that we will expand with more capabilities such as seamless upgrades in the future.

Done in `Our plans` and `Further work` paragraphs.

design/operator.md
```diff
@@ -0,0 +1,65 @@
+# Observerability
+
+* **Status**: PoC
+* **Author**: [slaskawi](https://github.com/slaskawi)
+* **JIRA**: [KEYCLOAK-10036](https://issues.jboss.org/browse/KEYCLOAK-10036)
+* **PoC video**: https://www.youtube.com/watch?v=nGsARoHk_Dw
+
+## Motivation
+
+* Simplify user experience on OpenShift
```
kfox1111: Is this intended to be restricted to just OpenShift? I would be interested in using it on Kubernetes as well.

design/operator.md
```diff
@@ -0,0 +1,65 @@
...
+
+The Operator can provision (and deprovision) RHSSO instances based on Keycloak CRD. The Operator will also spin up a Postgresql database and will create a Secret with an Admin credentials. It is also possible to attach different RHSSO plugins (such as metrics SPI plugin) and finalizers (that are responsible for cleaning up the resources (routes, PVs etc) allocated during provisioning phase).
+
+### Realm provisioning
+
+The Operator is also capable of provisioning RHSSO Realms. The KeycloakRealm CRD is very similar to what RHSSO exported JSON has. One of the most important things is that KeycloakRealm CRD contains information about clients.
+
+### Backups
+
+It is also possible to use a Cron Job for backing up a RHSSO installation. The backup definition needs to be put into Keyclaok CRD. Such definition contains an image responsible for backing up the database, encryption key and an AWS credentials secret name.
```
kfox1111: By AWS, do you mean s3?

design/operator.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Our plans
+
+We plan to take over the existing implementation that the Integreately Team have done (thank you!). Further work can be divided into stages:
+
+* [Stage 1:](https://issues.jboss.org/browse/KEYCLOAK-10318) Basic install capabilities:
+  * Transition Keyclaok Operator into Keycloak organization
```
kfox1111: s/Keyclaok/Keycloak/

design/operator.md
```diff
@@ -0,0 +1,65 @@
...
+  * Setup a simple testing with Travis
+  * Setup automatic deployments with Travis (or Jenkins)
+  * Tidy up the Readme file and implement basic documentation
+  * Publish the Operator to the Operator HUB
+  * Make sure everything work out of the box on Kubernetes
```
kfox1111: I think this answers my previous question but most of the document looks like it may be OpenShift only.

design/operator.md
```diff
@@ -0,0 +1,65 @@
...
+* [Stage 2:](https://issues.jboss.org/browse/KEYCLOAK-10319) Productization:
+  * Productization of Stage 1 (in other works, release RHSSO with all this bits)
```
kfox1111: s/works/words/ ?

design/operator.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Further work
+
+Here's a list of the features we're considering for the future:
+
+* Securing application automatically with Gatekeeper or Envoy JWT proxy
+* Clients as CRDs
```
kfox1111: +1. Maybe some code can be shared from https://github.com/kiwigrid/keycloak-client-controller/tree/master

design/operator.md
```diff
@@ -0,0 +1,65 @@
...
+* Seamless upgrades
```
kfox1111: One more I think would be interesting. To be able to split a Keycloak cluster over multiple k8s's for HA. So the Keycloak CR would have enough info to stand up its 1/3rd of the cluster.

kfox1111: Oh. And krb5 support. Storing the keytab in a secret and referencing it or some such api.

design/operator.md
slaskawi: That's for Kubernetes as well. Let me clarify this a bit more here.

design/operator.md
slaskawi: Yes.

design/operator.md
slaskawi: Good eye! Thanks for pointing it out.

design/operator.md
slaskawi: OpenShift (OKD) is basically Kubernetes plus additional features. So everything we'll implement here should (and will) work on vanilla Kube as well.

design/operator.md
slaskawi: Thanks for pointing this out!

design/operator.md
slaskawi: Thanks for sharing this!

The project looks interesting and we'll have a look into it. But just to be clear, our goal is the Operator Framework (and Operator SDK).

design/operator.md
slaskawi: Let me create a small paragraph about the ideas from the Community and put it there.

slaskawi:
> All in all, this looks to be a very good thing for Keycloak IMO.

@kfox1111 Thanks a lot for the review. I updated the document and re-pushed.

design/operator.md
kfox1111: Yeah. Having a crd that was similar so that ops could transition from one to the other would be nice. Sharing the code isn't needed.

PR #5, opened 2019-04-16
Multi factor and step up proposal

AlistairDoswald:
This design document is based on my [post in the mailing list on the administration for multi factor authentication](http://lists.jboss.org/pipermail/keycloak-dev/2019-March/011822.html). It has been amended from the following discussions, and has had it scope increased to include step-up authentication.

PR #11, opened 2019-06-07
W3C Web Authentication - Implementation stages - Authenticator only

tnorimat:
This design document treats the following three issues mentioned in [**W3C Web Authentication - Two-Factor**](https://github.com/keycloak/keycloak-community/blob/master/design/web-authn-two-factor.md) Implementation stages - Authenticator only

1. [KEYCLOAK-9360](https://issues.jboss.org/browse/KEYCLOAK-9360) : Two factor authentication with W3C Web Authentication
2. [KEYCLOAK-9358](https://issues.jboss.org/browse/KEYCLOAK-9358) : Investigate W3C Web Authentication libraries
3. [KEYCLOAK-9359](https://issues.jboss.org/browse/KEYCLOAK-9359) : Investigate how to test Web Authentication

### DIFF
--- a/design/web-authn-authenticator.md
+++ b/design/web-authn-authenticator.md
@@ -1,0 +1,429 @@
+# W3C Web Authentication - Implementation stages - Authenticator only
+
+* **Status**: Draft
+* **Author**: [tnorimat](https://github.com/tnorimat)
+* **JIRA**: [KEYCLOAK-9358](https://issues.jboss.org/browse/KEYCLOAK-9358), [KEYCLOAK-9359](https://issues.jboss.org/browse/KEYCLOAK-9359), [KEYCLOAK-9360](https://issues.jboss.org/browse/KEYCLOAK-9360)
+
+## Motivation
+
+As mentioned in [**W3C Web Authentication - Two-Factor**](https://github.com/keycloak/keycloak-community/blob/master/design/web-authn-two-factor.md), it is important to implement WebAuthn support to keycloak. This design document treats the following three issues mentioned in [**W3C Web Authentication - Two-Factor**](https://github.com/keycloak/keycloak-community/blob/master/design/web-authn-two-factor.md) Implementation stages - Authenticator only:
+
+1. [KEYCLOAK-9360](https://issues.jboss.org/browse/KEYCLOAK-9360) : Two factor authentication with W3C Web Authentication
+
+>  Acceptance criteria:
+>
+>    * Admin manually registers Web Authentication authenticator to replace OTP authenticator
+>    * Admin registers required action for user to register a security key
+>    * When user logs in again after registering the security key the user should be prompted to tap the security key after entering username and password
+
+2. [KEYCLOAK-9358](https://issues.jboss.org/browse/KEYCLOAK-9358) : Investigate W3C Web Authentication libraries
+
+>  The selected library should ideally support:
+>
+>    * MFA and passwordless
+>    * Username stored on device (no need to enter username when authenticating)
+>    * Attestation to be able to get details about authenticators and limit what authenticators should be supported by a realm
+>
+>  It should also have:
+>
+>    * A compatible open source license
+>    * A reasonable sized community
+>    * Good test coverage
+>    * Maven build (Graddle is ok, but not ideal)
+
+3. [KEYCLOAK-9359](https://issues.jboss.org/browse/KEYCLOAK-9359) : Investigate how to test Web Authentication
+
+## Relationship among other design documents
+***
+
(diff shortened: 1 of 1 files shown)


design/web-authn-authenticator.md
```diff
@@ -0,0 +1,429 @@
...
+## Relationship among other design documents
+***
+
+This design documents depends on the following two design documents:
+
+1. [**Application Initialted Actions**](https://github.com/keycloak/keycloak-community/blob/master/design/application-initiated-actions.md)
+
+It relates to how a user registers their public key credential on keycloak.
+
+Before realizing [**Application Initialted Actions**](https://github.com/keycloak/keycloak-community/blob/master/design/application-initiated-actions.md), an existing user having their user account can only register their public key credential in [User Account Service](https://www.keycloak.org/docs/latest/server_admin/index.html#_account-service).
```
stianst: Application Initiated Actions will be used by account console as well, so until this is added it won't be possible to register public key credential except with a required action at login time.

tnorimat: I see that point. I'll revise this design document including the following points:

* before realizing AIA, the user can register their own public key credential at their login time by Required Action.
* after realizing AIA, the user can also do it by AIA that User Account Service also use it.

Is that OK?

tnorimat:
I understand that only multiple public key credentials are supported in the new console and account endpoints.

Considering that multiple credential support is treated in https://github.com/keycloak/keycloak-community/pull/5,

* multiple public key credentials are supported after implementation of https://github.com/keycloak/keycloak-community/pull/5

* before it, multiple public key credential is not supported, only single public key credential is supported. But mention in this document that this single public key credential is tentative and omitted afterward.

Is that OK?

design/web-authn-authenticator.md
stianst: +1

tnorimat:
I've fixed the document as you pointed out (Application Initiated Actions, etc).

stianst:
I'm approving this as a draft. Will let @mposolda review and if he's happy he can merge it.

design/web-authn-authenticator.md
```diff
@@ -0,0 +1,425 @@
...
+
+* This setting is optional for WebAuthn API (`navigator.credentials.get()`). Therefore, if the switch if OFF, the executor of this WebAuthn API does not filter out authenticators by their user verification capability.
+
+### Specifying Public Key Credentials
+
+User can select which their public key credentials are used for their authentication.
+
+On user authentication:
+
+  * Keycloak shows the list of public key credentials with its metadata that the user has registered.
```
mposolda: Question: At stage 1, we will have just one public key per user. So would it makes sense, that if user has just one public key (which will be always at phase 1), the authenticator doesn't need to ask user, but automatically select that one public key? I guess it is not the case and it will be always good practice to ask user even if he has just one key, so that he confirms that he wants to use it?

I am not requesting to change anything, just asking to clarify this...

tnorimat:
I also like the option 2 as you said because I would like to reduce UI components as possible. I would like to revise this design document to show that we use ComboBox instead of the combination of Switch and Pull Down Menu (both only single item can be selected and multiple items can be selected). It it OK?

design/web-authn-authenticator.md
tnorimat: At stage 1, I will implement the features about Public Key Credential (e.g. credential selection on authentication)  the same as the later stage when multiple credentials are supported in order that make it easy to work on lager stages (e.g. multiple credentials per user supported ). Therefore, I will create the public key credential selection page on authentication time even if only single public key credential is supported at the stage 1. 

It seems to be quite odd for users that the features supposing multiple credentials are implemented at state 1. But as @stianst commented on this pull request, this single public key credential is tentative and not officially supported later on. Therefore, I think this implementation strategy at state 1 might be acceptable. WDYT?

If so, is it better to mention this in the design document?

mposolda: Thanks for the clarification. I agree to keep as it is and have the selection screen present even if there is just one key. Not sure whether to add a NOTE to this design, that selection will be useful especially in later stages or whether to keep design as is. I am fine with both options and leave to you.

mposolda:
@tnorimat Yes, So if you can revise this design document to remove some switches, I am completely fine with that. Thanks!

design/web-authn-authenticator.md
tnorimat: I'll add this notification on Public Key Credentials section.

tnorimat:
@mposolda I've revised this design document incorporating usage of ComboBox.

PR #16, opened 2019-07-01
Authentication session improvements (You are already logged-in)

mposolda:


design/authentication-flow-usability.md
```diff
@@ -0,0 +1,322 @@
...
+
+3. In our adapters, we may need to doublecheck if behaviour is correct and user will be redirected back to KC as described in
+the step 3 [above](#minimal-approach). For servlet application, I suppose it won't be an issue as redirect_uri is usually "secured" URI and hence automatically intercepted.
+I think it may be similar for other server adapters (node.js, gatekeeper etc) but will be good to doublecheck.
+For the keycloak.js, I think "login_required" will work fine and automatically redirect to KC, but for the applications with "check_sso" or without any "onLoad" value used,
+(For example _js-console_ application from the deprecated keycloak-examples), the adapter won't automatically redirect to Keycloak. So some changes in the adapter
+may be needed to achieve this.
+
+4. For SAML, will be good to doublecheck the behaviour. I suppose that SAML adapters are server-side and hence is not an
```
hmlnarik: No issue to redirect back for SAML secured area - the request would be resent unless anonymous access is also allowed into the secured area.

design/authentication-flow-usability.md
```diff
@@ -0,0 +1,322 @@
...
+anymore, but will be saved in the browser itself. This will be either in the request parameter sent across authentication requests or in the cookie, which will need to be "namespaced" to be specific per each browser tab. Few details to this:
+    
+    * The client data will be encoded in the JWS token. It can be pretty similar to the current KC_RESTART cookie - in other words, some JSON data signed with the realm secret key.
+    
+    * In ideal case, the token can be added as separate parameter to the authentication requests. In other words, instead of
+    parameters `code`, `execution`, `client_data` and `tab_id`, we will have `code`, `execution`, `tab_id` and `client_data_full` where
+    `client_data_full` will contain this encoded token. No need for `client_data` parameter, which has just few of client data.
+    
+    * There is a URL limit of 2000 characters (excluding host, but including the path, query params and fragment). So to be
+    on safe side, we can't include `client_data_full` as request parameter in case that is too long. I suggest to use 1200
```
hmlnarik: There is also limit on cookie size, number of cookies per domain, and total size of all cookies in a domain which varies across the browsers. This limit can be only 4096 bytes for the total size per http://browsercookielimits.squawky.net/. With all the info stored in the auth session notes, and some 10+ number of tabs, this limit could be reached quickly.

Since the limit is based on cookie domain and not cookie path, I don't think the boilerplate suffix would help here.

We thus need to find another way. Could we use POST instead of GET to pass the state? Perhaps part of the state could be stored in the URL (likely the `code`, `execution`, `tab_id`, `client_data` from minimal approach above)?

design/authentication-flow-usability.md
```diff
@@ -0,0 +1,322 @@
...
+
+    * In OAuth2/OIDC, there doesn't seem to be any nice error code for this purpose. See https://openid.net/specs/openid-connect-core-1_0.html#AuthError and https://tools.ietf.org/html/rfc6749#section-4.1.2. It seems best error code might be "server_error" and we may add the custom error_description, so that our adapters can understand it and handle it.
+
+3. In our adapters, we may need to doublecheck if behaviour is correct and user will be redirected back to KC as described in
+the step 3 [above](#minimal-approach). For servlet application, I suppose it won't be an issue as redirect_uri is usually "secured" URI and hence automatically intercepted.
+I think it may be similar for other server adapters (node.js, gatekeeper etc) but will be good to doublecheck.
+For the keycloak.js, I think "login_required" will work fine and automatically redirect to KC, but for the applications with "check_sso" or without any "onLoad" value used,
+(For example _js-console_ application from the deprecated keycloak-examples), the adapter won't automatically redirect to Keycloak. So some changes in the adapter
+may be needed to achieve this.
```
hmlnarik: Is checking adapters really needed? For the minimal approach, the worst what can happen is that user would have to click onto "login" button in the application which IMO is acceptable for the minimal solution.

mposolda: I was thinking about that before and IMHO the usability of that approach would be even worse than the usability of the current approach with showing the "You are already logged-in" page.

Currently when user is authenticated in tab1 and he fills the username/password in tab2, he is redirected to the "You are already logged-in" page. This is not ideal, but at least, user knows that something strange is going on. So he clicks on the "Back to application" and he would notice that he is not authenticated as he is aware of the strange/non-standard behaviour due the "You are already logged-in" page previously shown. 

With the approach you mention, after fill username/password screen, he will be redirected to the application like "js-console", but still in the anonymous mode. This can be quite confusing and user may think like "Why I am still not authenticated even if I just filled username/password" ? It can even have some security applications in theory (EG. user may not notice that he is not yet authenticated in the application and he may start to fill some forms in the application when still being in the anonymous mode etc).

That's the reason I propose to add some protocol-specific error parameters to the redirectUri to give adapters a chance to react on them and automatically redirect user back to the KC, which will do SSO dance and redirect user back as properly authenticated.

design/authentication-flow-usability.md
mposolda: So these apps, which allow anonymous access, is exactly my fear. If I understand correctly, after fill username/password in the browser tab2, user might be redirected back to the application, but in the anonymous mode. This can be quite confusing IMO as mentioned in the comment above. If we add some SAML specific errors, adapter can react on them and do the SSO dance instead of showing the anonymous page?

design/authentication-flow-usability.md
```diff
@@ -0,0 +1,322 @@
...
+2. Every page with the authenticators will contain some small JS code, which will periodically check the AUTH_TIMESTAMP cookie
+mentioned above. In case that it recognizes the change, the page will be automatically refreshed and moved to the latest state.
+For example if user confirms username/password in browser tab1, which moved to the TOTP screen, then tab2 will be moved in the
+background to the TOTP screen as well.
+In details, when javascript recognize the change in the timestamp cookie, then
+tab2 will:
+    * Send the browser "refresh" request
+    
+    * Authentication session will be shared per whole browser (with some exceptions described below in section [different flows in same browser](#different-flows-in-same-browser).
+    So on KC side, we will check the latest state from AuthenticationSession and update the page accordingly including URLs. This will allow all browser tabs will be updated to TOTP after confirm username/password in tab1
```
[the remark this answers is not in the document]
mposolda: I agree that first priority is to get rid of "You are already logged-in" . Also agree with the point of more uneccessary browser requests. However I think that manual refreshes is not very good. That would actually mean that user will in most cases still see the username/password screen and fill the username/password as most of users don't do browser refresh.

So maybe the JS can refresh the page in case that particular browser tab is clicked and gains focus? We can always refresh the page when gaining focus, which means that we won't need the timestamp cookie at all then.

The Idp flow or client-specific flow shouldn't be an issue. As mentioned below, there will be still more authentication sessions in the rootAuthSession to handle those cases. Example:
- There are 3 browser tabs. Tabs 1 and 2 are using flow1 (default realm flow). Tab 3 is using some client specific flow - flow2.
- User fills username/password on tab 1. He is moved to step 2 (TOTP screen)
- User clicks on tab 2. JS will refresh the page and user will be moved to TOTP screen
- User clicks on tab 3. JS will refresh the page, but in the flow 2 (client specific flow), user didn't anything yet. So refresh will just keep user at step 1 of the flow 2.

design/authentication-flow-usability.md
```diff
@@ -0,0 +1,322 @@
...
+    in tab1, KC will be able to lookup current authenticationSession and directly redirect user to the latest step. So for
+    example assuming that user is already on the TOTP screen in tab1 and he starts new authentication request to OIDC endpoint
+    in tab2, then he will be redirected to TOTP screen in tab2.
+    
+    * This also means that after authentication completed in tab1, KC redirects to the client application in particular tab as in
+    current Keycloak master. But in other browser tabs, there won't be immediate redirect to the applications in that case as this can have
+    some side-effects. For example unecessarily creating the HTTP session for the application where user was redirected in other browser tabs.
+    So in other browser tabs, we may display some "info" stating like "Authentication was completed. Redirecting to the
+    application in few seconds". There will be also some additional JS code, which will be triggered when user enters the particular browser tab. This JS
+    will refresh the page and will automatically redirect user to the application.
```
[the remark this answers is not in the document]
mposolda: Yes, kind of.

design/authentication-flow-usability.md
mposolda: Thanks for checking cookie sizes. That's not good news. My bad I did not check the cookie sizes myself :/ On the other hand, the http://browsercookielimits.squawky.net/ has latest update from 2013 and the most important browsers mentioned there has either no limit for domain or limit of 10234 (IE 8-15). Also the cookie is just a fallback and hopefully won't be needed in most of the cases.

It seems the options are:
- Don't do anything and not implement advanced design. Which means keep client data in the authSessions on the server. This can be done in case that minimal approach will address most of the use-cases and people won't create anymore usability JIRAs :) I fear this won't be the case and better usability may be needed.

- Count with the cookie limit based on the user-agent header. We will know the browser due the ua-parser recently added. In most of the cases, we won't either have limit or have bigger limit like 10234. So we will be able to keep just for example last 3 CLIENT_DATA_* cookies for last 3 shown browser tabs. For the other browser tabs, the redirect to the application will be needed as described in the "minimal" flow.

- Do something else like the POST approach. But I think that POST will be problematic. For example considering all the cases like browser buttons (back/forward/refresh). Currently we remove the POST request from browser history to ensure browser buttons work correctly. I can imagine POST can work just in case that browser is still sending GET requests, but POST requests are sent in the background through ajax. This will require more refactoring of authentication SPI itself to be trully JS based...

- Do something else and still keep the client data on the server rather than in the browser. I can think of something like divide current AuthenticationSession into 2 entities:
-- AuthenticationSession, which will keep just the authentication data
-- ClientAuthenticationSession, which will keep just the client data and reference to the AuthenticationSession. There will be ClientAuthenticationSession for each browser tab, but AuthenticationSession won't be necessarily specific to single browser tab.
However having things on the server doesn't properly solve the problem, that sessions on server can always expire. In case that session will be expired, redirect to the application will be still needed...

design/authentication-flow-usability.md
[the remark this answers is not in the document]
hmlnarik: > So maybe the JS can refresh the page in case that particular browser tab is clicked and gains focus? We can always refresh the page when gaining focus, which means that we won't need the timestamp cookie at all then.

+1 to this approach. When the user clicks the tab, we only need to put some notification to the user that the page is being refreshed so that they don't fill and submit a form in the meantime (could be just a shading full screen div or whatever)

design/authentication-flow-usability.md
hmlnarik: I'd rather add such a functionality only if really demanded by the community so that we keep the SAML adapter without Keycloak extensions and stick to the standard codes as much as possible. AFAIK, there is no standard SAML code that we could use.

mposolda: Ok, I am not so concerned about SAML TBH as I don't know how many people use the anonymous mode etc. For OIDC, I would add the error/error_description with some custom values, so that our adapters can react on it instead of showing anonymous page. I think that at least keycloak.js will need some support for this to be added. Simlarly I think that java servlet adapters don't need any change as redirect_uri is always "secured" URL and adapter always redirects to KC when redirect_uri is displayed. Not sure about gatekeeper and node.js

design/authentication-flow-usability.md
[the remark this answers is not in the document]
mposolda: I've changed the design per this.

pedroigor:
I was also wondering if server-side sessions could only be created after the user has provided at least his username (after the identification phase). I understand the reasons for using notes in session to store state during the login, but maybe we could avoid that if choosing @stianst suggestions.

mposolda:
Few points to this:

- Currently we defacto not share authentication session with multiple tabs. That's still true in the "Minimal design" . We just share the RootAuthSession, but that has just ID and list of attached authentication sessions. The main reason of RootAuthSession is the cookie AUTH_SESSION_ID, which provides session stickiness and ensure that infinispan entity is able to lookup on the current cluster node (performance improvement). The "minimal approach" just removes single authSession from the rootSession after authentication, but other authentication sessions (for other tabs) are still there. In reality, not much difference between the case when you have rootSession with the array of attached authSessions (not sharing any auth-state) or separate authSession per each tab. The only difference is the expiration (which is defacto shared per whole browser with rootAuthSessions).

- The main reason for "minimal approach" was to address 90% of the issues without a need of too much refactoring. That's what we discussed with @hmlnarik due the other priorities of team-puma. Minimal approach is just about removing single authSession from rootSession, and being able to redirect to client application and probably do something in adapters (especially in keycloak.js) to ensure that adapter is able to handle the redirection and redirect back to KC and automatically do SSO login. The part of redirect to client application and handle error on the adapter will be required in all the cases anyway (The "Advanced design" as well as your proposal), so I suggest to start with it and then see if it is sufficient based on the feedback from customers/community? Especially the "Minimal design" doesn't have any "Advanced" tricks. Especially no JS hacks involved in the authenticator pages, not additional cookies and not much refactoring in the authentication SPI. So maybe good place to start anyway as stage 1?

- Regarding, sharing the authSession, I think I agree with you that we may not need to share them among multiple browser tabs. And I agree that we may need some JS in the authenticator page, which will ensure that KC will be able to redirect automatically to the application when it detects that user was already authenticated in the other tab. But few more points to your point 4:

> Tab2 constantly monitors session cookie and detects session is created, which causes the page to be reloaded. 

IMO it will be better if instead of the constant monitoring, the JS event is triggered "onFocus". So just when particular browser tab is opened. In other words, when you have 5 browser tabs opened and you login in tab1, you won't automatically redirect also in 4 browser tabs directly, but just when the user explicitly clicks on the particular tab. That's what we discussed above with @hmlnarik.

> Authentication processing on server side detects login cookie and determines if authentication details in the login cookie is sufficient for what is request. If it is then redirects to app. If it is not then does whatever is needed to step-up the authentication

In your proposal, I am missing the details how you would detect that authentication details in the login cookie are sufficient? There are few things to consider like the flow type, flow ID (EG. Client-specific flow), step-up authentication, other OIDC/SAML parameters like "prompt", "max_age" etc. Also I am not sure I understand the purpose of the "revision" field on the session cookie and even the reason why the session cookie exists? I am not sure if it is anyhow easier than the "Advanced approach" and what benefits it will have over it?

- Regarding @pedroigor's comment on creating auth-session after username is known, this sounds like good idea, but I guess it will be more complicated to do. The devil lies in details IMO :) For example username/password authenticator may not be always the initial authenticator. The initial authenticator could be something else (For example WebAuthn with the passwordless flow) and the authenticator will require multiple steps and browser form submissions before it can identify user. So some server-side authentication state may be required even if user is not yet identified etc.

In shortcut, I would suggest to start with "Minimal design" for now and add more improvements (including JS support in authenticators, cookies etc) in some later stage...

PR #18, opened 2019-07-11
KEYCLOAK-3205 Vault SPI

hmlnarik:


design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+
+This is the list of vaults considered for a default implementations (TODO: to be shortlisted):
+
+* [Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/) / [OpenShift](https://docs.openshift.com/container-platform/4.1/nodes/pods/nodes-pods-secrets.html) secrets
+* Plain text (for testing purposes)
+* Database - use application level encryption to securely store secrets in the database
+  in a portable way. This would require dealing with keys as well as key rotation.
+* [HashiCorp vault](https://www.vaultproject.io/)
+* [Keywhiz](https://square.github.io/keywhiz/)
+* [Elytron Credential Store](https://docs.wildfly.org/17/WildFly_Elytron_Security.html#secure-credential-storage)
```
hmlnarik: Add section on where the credentials are stored / obtained at this time.

design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+     *        the secret.
+     * @param purpose Purpose of the secret, set by keycloak according to the caller context.
+     * @return The secret or {@code null} if the secret was successfully resolved, or
+     *         an empty {@link java.util.Optional}
+     */
+    Optional<byte[]> obtainSecret(KeycloakContext context, String vaultSecretId, String purpose);
+}
+```
+
+Vault configuration has to contain mapping between purpose and the allowed provider IDs sorted by priority to be used for this purpose, e.g. `realm-key => HSM, vault-1`, `smtp => vault-1`, `default => vault-2, HSM` would mean that:
```
slaskawi: I have some doubts how this approach will scale if we add more features to Keycloak. Note, that every time you add something new to Keycloak that requires storing some sort of credentials, you will need to revisit this mapping and add new parameters to it.

How about doing this differently. Imagine, that you're storing SMTP credentials and you'd like to store both, username and password in a vault. A Keycloak Admin could do it in the Email configuration screen by setting username equal to `${vault.vault-1.smtp.username}` and `${vault.vault-1.smtp.username}`. The Keycloak could extract from this string everything that is needed to to grab this value from a vault, namely:
- Vault id is `vault-1`
- Vault secret id is `smtp.username`

When a realm is loaded, the Keycloak implementation will scan all the fields and if `${vault.*}` is found, use the `VaultServiceProvider` (passing both vault id and vault secret id) to obtain proper values. 

Note, that this implementation doesn't require hardcoded mapping and is very scalable. It can be easily written in a generic way.

design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+     *
+     * @param context Context that contains info about current request, to provide e.g. realm separation
+     * @param vaultSecretId Identifier of the secret. It corresponds to the value entered by user
+     *        in the respective configuration, which in turn is obtained from the vault when storing
+     *        the secret.
+     * @param purpose Purpose of the secret, set by keycloak according to the caller context.
+     * @return The secret or {@code null} if the secret was successfully resolved, or
+     *         an empty {@link java.util.Optional}
+     */
+    Optional<byte[]> obtainSecret(KeycloakContext context, String vaultSecretId);
```
slaskawi: I believe we could put `vaultId` argument here, instead of `purpose` in the `VaultProvider`.

design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+}
+```
+
+#### Default vault service providers to consider
+
+This is the list of vaults considered for a default implementations (TODO: to be shortlisted):
+
+* [Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/) / [OpenShift](https://docs.openshift.com/container-platform/4.1/nodes/pods/nodes-pods-secrets.html) secrets
```
slaskawi: This one can be interpreted as a `paint text` vault. Kube can mount a secret as a key-value-based file.

design/secure-credentials-store.md
hmlnarik: > I have some doubts how this approach will scale if we add more features to Keycloak. Note, that every time you add something new to Keycloak that requires storing some sort of credentials, you will need to revisit this mapping and add new parameters to it.

No, this is auxiliary class with constants that need not be used but is beneficial. One can still use plain `String` when desired. However, it is better to have compiler check that we use proper context, compiler would not fail for the following:

```java
 sp.obtainSecret(context, "vault-id", "stmp");
```
but would fail for
```java
 sp.obtainSecret(context, "vault-id", Purpose.STMP_SECRET);
```
while it is still possible to write:
```java
 sp.obtainSecret(context, "vault-id", "smtp");
```
but rather error-prone. Why would not this scale?

> A Keycloak Admin could do it in the Email configuration screen by setting username equal to `${vault.vault-1.smtp.username}` [...]

The drawback of the approach with parsed `vault.vault-1.smtp.username` is that `vault-1` allows for no specification of the purpose and uses strictly that vault for `smtp.username` secret. Changing vaults would not be possible with this approaches without changing the entered properties.

design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+2. Encryption / decryption SPI for secrets. There are services providing encryption / decryption
+   of secrets, while abstracting from key management and other stuff.
+
+The rest of this document will deal with the first part (Vault SPI) only. **The second part
+(encryption / decryption) will be treated in a separate document.**
+
+
+## Scope
+
+Generally any secret can be stored in a vault. The value stored in Keycloak database should
```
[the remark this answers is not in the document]
hmlnarik: Fixed

design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+It may be beneficial to allow per-type configuration. For example retrieving LDAP and SMTP credentials from Kubernetes secrets, while using HashiCorp Vault Transit Secrets for external tokens and not encrypting user secrets like otp secrets. That would allow more flexiblity as well as tuning security vs performance.
+
+Vault will manage access to vault providers and ensure proper vault is used to obtain the secret.
+
+For the consumer, the interface would be like this:
+
+```java
+public interface VaultProvider {
+
+    public static class Purpose {
```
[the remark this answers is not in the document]
slaskawi: Good point @hmlnarik. I haven't thought about binary compatibility at the API level.

design/secure-credentials-store.md
```diff
@@ -21,51 +21,119 @@ Credentials include:
...
+
+### Realm Keys
+
+Today there are SPIs both for realm keys as well as signing of tokens. Keys SPI enables storing keys in Java KeyStore as well as adding custom support for a vault. Signature SPI enables delegating token signature to external HSM devices.
+
+## Design
+
+### Vault Provider SPI
+
+It may be beneficial to allow per-type configuration. For example retrieving LDAP and SMTP credentials from Kubernetes secrets, while using HashiCorp Vault Transit Secrets for external tokens and not encrypting user secrets like otp secrets. That would allow more flexiblity as well as tuning security vs performance.
```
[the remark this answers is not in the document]
hmlnarik: Apparently my yesterday's reply to this comment has been lost, sorry about that.

In short - I believe we should not. Vault (`VaultProvider`) is an aggregation of named vault service providers (`VaultServiceProvider`). `VaultServiceProvider` is an implementation of a specific vault, thus it should not be aware of the its vault ID, which is only an id of the instance within `VaultProvider`.

For example, some secrets (e.g. LDAP, realm private keys) might need to be stored very securely (expensive HSM with small capacity), the rest of secrets might be stored elsewhere (software vault). As an administrator, one can specify the secrets for given purpose (realm keys, LDAP) should be retrieved from specific vault(s) instead of hardcoding the vault ID in a property name. Furthermore, keycloak realm administrator need not even be aware of the existence of two or more different vault types.

Purpose maps to several vaults, which allows easy transition from one vault to another (e.g. from software to HSM), if the secret IDs can stay the same (e.g. `${vault.realm-key}`)

design/secure-credentials-store.md
stianst: Good point. The Kubernetes provider would just read values from a file per secret, or from an environment value. That means the plain text is not needed, but rather the Kube/OpenShift one can just be used for testing purposes.

design/secure-credentials-store.md
```diff
@@ -21,51 +21,120 @@ Credentials include:
...
+
+This is the list of vaults considered for a default implementations (TODO: to be shortlisted):
+
+* [Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/) / [OpenShift](https://docs.openshift.com/container-platform/4.1/nodes/pods/nodes-pods-secrets.html) secrets
+* Plain text (for testing purposes)
+* Database - use application level encryption to securely store secrets in the database
+  in a portable way. This would require dealing with keys as well as key rotation.
+* [HashiCorp vault](https://www.vaultproject.io/)
+* [Keywhiz](https://square.github.io/keywhiz/)
+* [Elytron Credential Store](https://docs.wildfly.org/17/WildFly_Elytron_Security.html#secure-credential-storage)
```
stianst: What does Elytron actually support? Is it just a basic vault, or does it delegate to others? If just a basic vault it should be dropped from being considered, but if it enables support for multiple vault providers itself, than it could make a lot of sense.

design/secure-credentials-store.md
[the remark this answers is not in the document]
slaskawi: > In short - I believe we should not. Vault (VaultProvider) is an aggregation of named vault service providers (VaultServiceProvider). VaultServiceProvider is an implementation of a specific vault, thus it should not be aware of the its vault ID, which is only an id of the instance within VaultProvider.

@hmlnarik Thanks. This clarifies it a bit. I though the hierarchy is the other way around. Could we put that explanation somewhere in the design?

> Furthermore, keycloak realm administrator need not even be aware of the existence of two or more different vault types.

@hmlnarik I've always assumed, that Keycloak Admin is the same persona that configures Keycloak though the admin console and configures the deployment (e.g. configuring the SPIs in `standalone.xml`). 

> Purpose maps to several vaults, which allows easy transition from one vault to another (e.g. from software to HSM), if the secret IDs can stay the same

@hmlnarik Ok, thanks for the explanation

design/secure-credentials-store.md
slaskawi: > Why would not this scale?

@hmlnarik I'm looking at this problem from the development perspective. Let's assume that we would like to introduce a new feature that uses some sort of credentials (that we want to keep confidential, thus we want to use Vault to store them). Implementing such a new feature would require adding a new mapping, e.g. `myfeature => vault-5`. This mapping will be interpreted by some piece of Java code. This means, that every new feature that wants to use Vault, would need to modify that piece of Java code and add its own mapping there.

My proposal (described the above), doesn't require modifying such a mapping. You could use Vault out of the box anywhere in the configuration. You just need to follow some sort of naming convention. So for a `myfeature` example, I would just need to configure credentials in the Admin Console like this: `${vault.vault-5.myfeature.credentials}`.

Just to clear it even more, the whole interaction across Keycloak would look like this (it's a bottom up point of view):

```
MyFeaureModelProvider
     | <-- Here's where we query for ${vault.vault5.myfeature.credentials}
     | <-- and obtain de-vaulted credentials
VaultProvider (has a map of all vaults)
     |
VaultServiceProvider (implements MyVault)
     |
MyFeatureModel (an Entity used by MyFeature)
     |
Hibernate
     |
Database
```

So in this context, whenever I mention it "scales", I mean, "how easy it is to add a new feature with credentials stored in a vault".

design/secure-credentials-store.md
slaskawi: As far as I know, this is just a standalone vault that uses files. It doesn't allow us to delegate to other vaults.

slaskawi:
> We don't need to support multiple vaults and especially not a vault per-type. A single vault per-server is more than sufficient for now.

> We should also add a section regards to isolating secrets from different realms.

@stianst I see some clash between those two points. If we had a single vault per Keycloak server, this means it will be shared for all the realms. So, effectively there will be no isolation at all. 

@hmlnarik @stianst What if we designed it as a Vault per Realm instead of Vault per Keycloak Server? I believe such a design might be a bit more flexible. The most interesting aspect would be how to express a binding between a particular Vault and a particular Realm. This mapping would need to be somehow dynamic (so that it wouldn't require a server restart to update it).

slaskawi:
@hmlnarik Thanks for clarification! I believe I'm done with this round of review. Once you have anything new, please let me know. I would be more than happy to visit this PR.

design/secure-credentials-store.md
hmlnarik: To sum up an offline discussion - if desired, all of this could be done inside a special vault provider instead of a built-in solution that would mandate a specific property naming.

design/secure-credentials-store.md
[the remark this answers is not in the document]
hmlnarik: Let's keep things simple then. There will be no purpose and there will be only a single vault per server for the initial implementation. If there would be a need for more, we can either add purpose into Keycloak context, or extend the interface later with default implementation.

design/secure-credentials-store.md
hmlnarik: Addressed.

hmlnarik:
Merging as draft after addressing the first round of comments.

PR #21, opened 2019-08-16
User Event SPI

vmuzikar:


design/user-event-spi.md
```diff
@@ -0,0 +1,73 @@
...
+
+Key changes over current generic Event SPI:
+* Each event is fired twice – as a pre- and post-event.
+* Pre-events are "interactive" – listener can approve, reject or delegate (for later decision by an external system)
+  each pre-event. 
+* The same User Event SPI will be used everywhere – admin console and REST endpoints, user registration form,
+  account console and REST endpoints, etc.
+* The event types will be much more detailed, not just usual CRUD. There'll be separated events for user self-registration,
+  creation through Admin REST API, Account Console, etc.
+
```
slaskawi: I would love to see some prototype written as code here. How the interface would look like. What would be the events definition etc.

design/user-event-spi.md
```diff
@@ -0,0 +1,73 @@
...
+### Terminology
+
+User Action (in this context) is a process which fires the User Event. This can be e.g. a user creation through Admin REST
+API, user self-registration, user field modification, assigning a user role, etc.
+
+Pre-event is an event that is fired after all other verifications passed (authorization, request validation, etc.), just
+before persisting any changes.
```
slaskawi: Hooking in here is a bit tricky as we will probably be inside a JTA transaction. Having a long-running code here or communicating with an external system might be dangerous. 

My intuition here is that a universal design would require some kind of persistent queues somewhere (as we probably would like those approvals to survive restarts).

design/user-event-spi.md
```diff
@@ -0,0 +1,73 @@
...
+
+**When an event is rejected:**
+* The SPI throws an exception.
+* Based on this exception the User Action is terminated and some meaningful error message is sent.
+
+**When an event is delegated:**
+* The SPI stores any necessary data (representations, etc.) required later in case this event is approved.
+* The SPI throws an exception.
+* Based on this exception the User Action is intercepted and some meaningful error message is sent.
+* A one-time token is created. The token is used later by an external system to approve or reject this event.
```
slaskawi: What about clustering here? Imagine that someone edits his email and we generate a one-time-token to get an answer from an external system. What would happen is Keycloak consists of 2 servers behind a load-balancer (a typical example of Keycloak deployment), where the first one issued a token and a second one received an approval?

design/user-event-spi.md
vmuzikar: Sure, I plan to add some list of the new events as well as some code snippets (probably on Friday).

design/user-event-spi.md
vmuzikar: Yes, it's really tricky. But I'd like to use the idea from the [Approvals System PoC](https://github.com/keycloak/keycloak-community/blob/master/design/approvals-system.md) – afterall it was @stianst's idea to transform the PoC to new User Event SPI. :)

design/user-event-spi.md
vmuzikar: That's a good question and honestly I don't have the answer yet.
But it doesn't need to be anything over sophisticated – it's just a "regular" token quite similar to e.g. Initial Access Token when registering clients. And it somehow works with clustering.

vmuzikar:
@slaskawi Thank you very much for your valuable feedback!

design/user-event-spi.md
slaskawi: I'm absolutely for it. I think this SPI is a good idea. 

I would just like to clarify, what happens if:
* We have pending approvals (or in-flight pre-events) and we restart Keycloak. I can easily see this happening in Kube.
* We want to send a pre-event for evaluation for an external system and that system is offline at the moment or responses very slowly (e.g. 1 response per 10 minutes, which usually exceeds JTA timeout).

design/user-event-spi.md
vmuzikar: @slaskawi I've added the list of event types as well as some code snippets. Those are really high-level and from the perspective of a Keycloak dev.

design/user-event-spi.md
vmuzikar: >We have pending approvals (or in-flight pre-events) and we restart Keycloak. I can easily see this happening in Kube.

Pending approvals are persisted. It's not just in memory.

>We want to send a pre-event for evaluation for an external system and that system is offline at the moment or responses very slowly (e.g. 1 response per 10 minutes, which usually exceeds JTA timeout).

That's a good question. There could be some fail safe mechanism that would iteratively retry it as a background task (the event is either way marked as delegated). But IMHO we should rather fail the operation completely.

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+### Event types
+
+* User is created manually through admin endpoints
+* Basic user info (name, email, etc.) is modified through admin endpoints
+* Basic user info (name, email, etc.) is modified through account endpoints
+* User self-registers through login page
+* User self-registers through an IdP, i.e. logs in for the first time
+* Admin adds a federation link
+* Admin removes a federation link
+* User logs in using federation for the first time
```
jonm-bb: How about when the user is created via federation? E.g. when a user is added to the company LDAP, it is synchronised to the relevant Keycloak Realms. From there, the Event SPI could be used to synchronise to another downstream system.
This would be very useful for syncing users to other systems, e.g. a HR system

siepkes: > From there, the Event SPI could be used to synchronise to another downstream system.
This would be very useful for syncing users to other systems, e.g. a HR system

Isn't that something that belongs in an IDM system (Midpoint IDM, Wren:IDM, etc.) instead of an AM system like Keycloak?

vmuzikar: @jonm-bb Technically speaking it'd be possible. The use case you're describing is exactly what I had in mind by "User logs in using federation for the first time". In other words, let's say a user exists in some LDAP; then the user logs in for the first time to Keycloak (i.e. user is created in Keycloak via federation); then an event is fired.

jonm-bb: @siepkes - sure, you could involve more products and systems, but I think this kind of SPI makes integrating with such systems easier, and could even avoid then need for them for simple deployments. If it doesn't add significantly to the complexity, then it would add quite a lot of value to the product.
@vmuzikar - ah cool, makes sense. Could we clarify that the event is fired when the user is persisted as well as when they log in?

design/user-event-spi.md
slaskawi: Thanks!

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+            UpdateBasicInfoUserPreEvent updateBasicInfoEvent = (UpdateBasicInfoUserEvent) event;
+            String current = updateBasicInfoEvent.getCurrent().getEmail();
+            String suggested = updateBasicInfoEvent.getSuggested().getEmail();
+            if (!current.equals(suggested)) {
+                String token = event.delegate("Email changes require admin approval");
+                startBPMSProcess(event, token);
+            }
+        }
+        else {
+            event.approve();
```
slaskawi: This one looks interesting. What would happen if I forgot to write the `event.approve()`? I assume that it would become approved by default?

design/user-event-spi.md
slaskawi: Or maybe there should be way to get all failed events somehow (and a user would need to implement a small piece of code to retry or remove them)? If there would be a way to do that - I'm totally fine!

design/user-event-spi.md
slaskawi: Ok, if you checked that and it's working fine - I'm good.

design/user-event-spi.md
vmuzikar: Or perhaps the whole mechanism for sending pre-events to external system could be async somehow. This would minimize any performance impact. All the User Action needs to know is that an approval request was delegated – when that actually happens is not that important.

design/user-event-spi.md
vmuzikar: Well, I haven't checked. :) I just suppose so because tokens and clustering is not a new thing.

design/user-event-spi.md
vmuzikar: @jonm-bb Current design includes only events for when the user is persisted.

design/user-event-spi.md
vmuzikar: I'd say something like `IllegalStateException` would be thrown.
But from my PoV this is just a minor detail – it could be implicitly approved as well.

vmuzikar:
@pedroigor Thank you for pointing this out! I'll look into it and modify the design accordingly.

design/user-event-spi.md
jonm-bb: > Current design includes only events for when the user is persisted.
Okay, then so this line:
> User logs in using federation for the first time
I think we could change this to "User record is created via federation"
Does that make sense?

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+* User is removed from a group (through admin endpoints)
+* User has a realm or client role assigned (through admin endpoints)
+* User has a realm or client role unassigned (through admin endpoints)
+* Admin updates a user's password through admin endpoints (not interactive)
+* User updates their password (not interactive)
+* User adds an authenticator
+* User removes an authenticator
+* Admin removes an authenticator from a user
+* User is deleted
+
```
velias: I miss few event types there we already use in our keycloak extension:
* auth flow initiates for user (keycloak obtained auth request by protocols like OIDC or SAML2 and starts auth flow)
* auth flow is finished for user (just before user's browser is redirected back to client app with OIDC or SAML2 etc)
* user performed successful authentication (by username/password form submit, social login or any other authenticator like SMS) - may be fired multiple times during auth flow if more factor auth is used
* required action performed/finished by user (including email verification, save of updated user info etc) - may be fired multiple times during auth flow if more required actions is execured

stianst: The idea is to split the current event listener into multiple more bespoke apis. This one is for user crud operations. We'd add a different one for login/actions/etc.

velias: I see, that's true that we do not need pre and post for events I proposed, thanks

vmuzikar:
@pedroigor I've studied [RFC 8417](https://tools.ietf.org/html/rfc8417). It's basically just a specialized JWT and even though it covers event related messages, I don't think it should be the main "communication protocol" for User Event SPI.

The specs states:
>While SET [Security Event Token] recipients will often take actions upon receiving SETs, security events cannot be assumed to be commands or requests.

This is from my PoV is a bit conflicting with the Interactive pre-events concept, as each pre-event is sort of "approval request" (of course not always – when there's no Listener interested).

Also, limiting the external communication to SET could introduce more complexity, like e.g. adapters. Let's say I want to approve selected pre-events through a business process in BPMS, so I need to start the process for each pre-event that I'm interested in.
If SET would be used, I would need to implement some sort of adapter (external app) that would receive SETs and then started the business process using BPMS REST API.
Without SETs, I could start the process directly from a custom User Event Listener (could be done async as @slaskawi suggested).

So I believe if we were to use SETs, it would have to be in form of a separate User Event Listener that would offer configurable recipients, but it wouldn't support the events interception, i.e. approval mechanics. However, this is from my PoV out of scope of this design proposal (which doesn't cover specific implementation of various Listeners).

@pedroigor Makes any sense?

stianst:
+1 With regards to RFC 8417. We should have our own webhook rest API, but can also consider to support RFC 8417 in addition if needed.

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+before persisting any changes.
+
+Post-event is an event that is fired after a User Action has finished completely and all changes are persisted.
+
+### Event types
+
+* User is created manually through admin endpoints
+* Basic user info (name, email, etc.) is modified through admin endpoints
+* Basic user info (name, email, etc.) is modified through account endpoints
+* User self-registers through login page
```
driverpt: I would just add the event type of `UserCreated`. And then add a `source` field inside.

From a Microservices perspective, in like 90% of the cases, it's irrelevant where user came from, so the `source` should be a complementary field and not a specific event

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+  before the new account is actually fully created.
+* A user self-edits their email address. The change must be approved first by an external system before it's actually
+  applied.
+
+## Design proposal
+
+To overcome the current Event SPI limitations, a new "domain-specific SPI" will be implemented – User Event SPI.
+
+Key changes over current generic Event SPI:
+* Each event is fired twice – as a pre- and post-event.
```
driverpt: I think we're mixing concepts. I would split this into 2.
 - Event SPI
 - Interceptor SPI

One thing is listening to events, a different thing is to intercept and modify them

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+* Pre-events are "interactive" – listener can approve, reject or delegate (for later decision by an external system)
```
driverpt: I would remove the `approve`, `reject` or `delegate`. If something goes wrong, just throw an Exception or `revert` the event

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+  each pre-event. 
+* The same User Event SPI will be used everywhere – admin console and REST endpoints, user registration form,
+  account console and REST endpoints, etc.
+* The event types will be much more detailed, not just usual CRUD. There'll be separated events for user self-registration,
```
driverpt: I disagree on this. When a new user is created in Keycloak, a `UserCreated` Event should be fired, regardless of `self-registration` or `Federated-Login` or `Social-Provider`. Just add a `Source` object or field with that complementary information.

design/user-event-spi.md
```diff
@@ -0,0 +1,149 @@
...
+## Use cases
+
+* A user is removed from Keycloak. All related records in an external system are removed before the user is permanently
+  deleted (e.g. eshop orders).
+* A user self-registers in Keycloak. Their registration must be approved by an external system (either human or automated)
+  before the new account is actually fully created.
+* A user self-edits their email address. The change must be approved first by an external system before it's actually
+  applied.
+
+## Design proposal
```
driverpt: There's one thing missing here in the Design Proposal. The Event dispatcher should be run on a separate and independent thread so it doesn't block Keycloak Operations

stianst:
Should consider cloud events

stianst:
https://github.com/cloudevents/sdk-java

PR #39, opened 2019-10-22
Updated multi-factor-admin-and-step-up document to draft 3

AlistairDoswald:
The modifications reflect the state of the prototype, they include:
- Increasing the description of conditional flows and conditional executions
- A more important description of the "back" functionality
- Replacing ascii art depictions by screenshots from the prototype in most places
- Correcting examples to the actual names and behaviour that would be encountered
- Adding an example flow for conditional flows
- Correcting code examples to the actual code used in the prototype
- Correcting the default flows to the actual default flows in the prototype
- Moving the description of changes of required actions to the "Future enhancements" appendix
- Adding screen shots to show the selection of credentials by users during the authentication
- Adding screen shots to show the administration of credentials
- Removed the "Authentication Flow examples" appendix. The problems they showed can no longer happen in the current design of the prototype.
In addition, there are a few minor modifications to the text and to the formatting.


design/multi-factor-admin-and-step-up.md
```diff
@@ -171,145 +160,117 @@ For multi-token support in keycloak, when an authentication execution uses crede
...
-Auth type                         | Requirement
------------------------------------------------------------------------------------------------
-Username Validation                [ ] Alternative  [x] Required                   [ ] Disabled
-Password                           [ ] Alternative  [x] Required                   [ ] Disabled
-2nd Factor                         [ ] Alternative  [x] Required  [ ] Conditional  [ ] Disabled
-  | - OTP                          [x] Alternative  [ ] Required                   [ ] Disabled
-  | - Other                        [x] Alternative  [ ] Required                   [ ] Disabled
-```
+For the direct grant flow:
+![Default Direct Grant](img/multi-factor-admin-and-step-up/DefaultBrowser.png)
```
mposolda: The screenshot for "Direct Grant" contains browser flow. Looks like a mistake?

design/multi-factor-admin-and-step-up.md
```diff
@@ -171,145 +160,117 @@ For multi-token support in keycloak, when an authentication execution uses crede
...
+    return (OTPCredentialProvider)session.getProvider(CredentialProvider.class, "keycloak-otp");
 }
 ```
-Having the associated credential provider directly in the class allows the retrieval of the credentials through the `UserCredentialManager` and the type given by the `CredentialProvider`. It also simplifies the validation code a little, as it's no longer necessary to pass by the `UserCredentialManager` to call `isValid()`, but it can be called directly from the associated `CredentialProvider`.
-
-The `Authenticator#configuredFor()` method may become obsolete with this scheme, as it is superseded by the fact that a class extends `CredentialValidator` and the list of credentials that is obtained for the user.
+Having the associated credential provider directly in the class allows the retrieval of the credentials through the `UserCredentialManager` and the type given by the `CredentialProvider`. It also simplifies the validation code a little, as it's no longer necessary to pass by the `UserCredentialManager` to call `isValid()`, but it can (and should) be called directly from the associated `CredentialProvider`.
```
mposolda: For this particular part, I personally want to doublecheck the prototype a bit more and verify if it doesn't bring any regression with relation to userStorage providers. AFAIK in current Keycloak master, it is possible to specify that your userStorage provider is able to validate certain credential types (password, otp). I am not 100% sure if this change mentioned in this paragraph and introduced in OTPAuthenticator won't introduce any limits/regressions for the usage with userStorage provider, but I hope it is ok...

I am fine with that design for now and will get back to you if I have some doubts about userStorage providers. Just wanted to let you know.

design/multi-factor-admin-and-step-up.md
AlistairDoswald: yes, I've pushed a correction

mposolda:
Merging, the design is not 100% aligned with what we have in keycloak master after all, but it is better than "draft 2" design, which is much more out-of-date with regards to the latest master changes.

PR #44, opened 2020-01-15
KEYCLOAK-11612 Client conformance profiles

tnorimat:
This design document treats the following issue.

1. [KEYCLOAK-11612](https://issues.jboss.org/browse/KEYCLOAK-11612) : [SPIKE] Client conformance profiles

### DIFF
--- a/design/client_conformance_profiles.md
+++ b/design/client_conformance_profiles.md
@@ -1,0 +1,550 @@
+# Client Conformance Profiles
+
+* **Status**: Draft #1
+* **JIRA**: [KEYCLOAK-11612](https://issues.jboss.org/browse/KEYCLOAK-11612)
+
+## Motivation
+
+To protect APIs that require high level security like providing financial services, there are several API security profiles like Financial-grade API Security Profile.
+
+To enforce the client operating in complying with such the security profile, the client conformance profiles are introduced. 
+
+## Overview
+
+The client conformance profile's tasks are the following.
+
+* Client's Configuration Check
+
+When registering the newly client or configuring the client that has already been registered, the client conformance profile checks whether the configuration of the client complies with the applied security profile.
+
+If not, keycloak rejects this registration/configuration and returns the appropriate error.
+
+* Client's Request Check
+
+When the client sends the request, the client conformance profile checks whether this request complies with the applied security profile.
+
+If not, keycloak rejects this request and returns the appropriate error.
+
+* Security Profile Specific Operation on the Client's Request
+
+Depending on the security profile, keycloak needs to do this profile specific operation when it receives the request from the client.
+
+## Scope
+
+### Target Security Profile
+
+The following security profiles are to be supported.
+
+* [FAPI (Financial-grade API) Read Only API Security Profile](
+https://openid.net/specs/openid-financial-api-part-1-ID2.html)
+* [FAPI (Financial-grade API) Read and Write API Security Profile](https://openid.net/specs/openid-financial-api-part-2-ID2.html)
+
+The following FAPI based security profiles might be supported in the future.
+
+* UK : [OpenBanking Security Profile](
+https://openbanking.atlassian.net/wiki/spaces/DZ/pages/7046134/Open+Banking+Security+Profile+-+Implementer+s+Draft+v1.1.0)
+* Australia : [CDR (Consumer Data Rights) Security Profile](
+https://consumerdatastandardsaustralia.github.io/standards/#security-profile)
+
+
+## Design
+
+### Scope of client conformance profile
+
+Keycloak can select and set the client conformance profile per client.
+
+### The number of applicable client conformance profile simultaneously
+
+Keycloak can set only one client conformance profile to the client.
+
+### When and how keycloak selects and sets the client conformance profile to the client
+
+The administrator must be involved in the process of determining and setting the client conformance profile to the client.
+
+It must be prohibited for the client itself to determine and set the client conformance profile by itself, without the administrator's involvement.
+
+The ways of client registration are as follows.
+
+* Dynamic Client Registration Endpoint
+    - Keycloak Client Representation
+    - OpenID Connect Client Metadata Description
+
+* Client Registration CLI
+
+* Admin Console
+    - UI
+    - CLI
+    - REST API directly
+    - Import
+ 
+The following 2 ways of client registrations can only determine and set the client conformance profile to the client.
+
+* Dynamic Client Registration Endpoint
+    - OpenID Connect Client Metadata Description
+        * Use Initial Access Token
+
+When creating the initial access token on the admin console, keycloak provides the pull-down menu for specifying the client conformance profile. The selected client conformance profile is included onto the initial access token to convey it to dynamic client registration endpoint afterwards.
+
+* Admin Console
+    - UI
+
+When creating the client on the admin console, keycloak provides the pull-down menu for specifying the client conformance profile.
+
+### When and how keycloak changes the client conformance profile to the client that has already been registered
+
+The administrator must be involved in the process of changing the client conformance profile to the client.
+
+It must be prohibited for the client itself to change the client conformance profile by itself, without the administrator's involvement.
+
+The ways of client configuration are as follows.
+
+* Dynamic Client Registration Endpoint
+    - Keycloak Client Representation
+    - OpenID Connect Client Metadata Description
+
+* Client Registration CLI
+
+* Admin Console
+    - UI
+    - CLI
+    - REST API directly
+    - Import
+
+The following 1 way of client configuration can change the client conformance profile to the client.
+
+* Admin Console
+    - UI
+
+### Client's Configuration Check
+
+Keycloak needs to do Client's Configuration Check at the following points.
+
+* Dynamic Client Registration Endpoint
+    - Keycloak Client Representation
+    - OpenID Connect Client Metadata Description
+
+* Client Registration CLI
+
+* Admin Console
+    - UI
+    - CLI
+    - REST API directly
+    - Import
+
+### Client's Request Check
+
+Keycloak needs to do Client's Request Check at the following points.
+
+* Authorization Request
+* Token Request
+* Token Refresh Request
+* Token Introspection Request
+* UserInfo Request
+* (Back Channel) Logout Request
+
+### Security Profile Specific Operation on the Client's Request
+
+Depending on applied security profile, keycloak needs to do Security Profile Specific Operation on the Client's Request at the following points.
+
+* Authorization Response
+* Token Response
+* Token Refresh Response
+* Token Introspection Response
+* UserInfo Response
+* (Back Channel) Logout Response
+
+### Client Conformance Profiles SPI
+
+Depending on applied security profile, keycloak needs to process Client's Configuration Check, Client's Request Check, and Security Profile Specific Operation on the Client's Request.
+
+To realize it, Client Conformance Profile SPI is introduced. Its provider implementations are also provided per client conformance profile. 
+
+## Implementation
+
+### Plan
+
+Ph 1. Introducing Client Conformance Profile Mechanism (SPI Provider)
+
+Ph 2. Supporting FAPI Read Only Security Profile (Implementing Provider)
+
+Ph 3. Supporting FAPI Read and Write Security Profile (Implementing Provider)
+
+Ph (TBD). Supporting OpenBanking Security Profile
+
+Ph (TBD). Supporting CDR Security Profile
+
+### FAPI Read Only API Security Profile
+
+#### Endpoint : Dynamic Client Registration Endpoint
+
+Client's Configuration Check
+
+* Operations :
+    - Registering the client
+    - Configuring the existing client
+
+##### 1. Client Authentication Method
+OpenID Connect Client Metadata - `token_endpoint_auth_method` is one of the following values.
+ 
+ 1. `tls_client_auth_subject_dn` (currently, keycloak only supports it for Mutual TLS for OAuth Client Authentication[MTLS])
+ 2. `client_secret_jwt`
+ 3. `private_key_jwt`
+
+* Reference:
+    - Authz Server / 4, 4-1, 4-2
+    - Confidential Client / 1, 1-1, 1-2
+
+Error Response : 400 Bad Request (the detail is TBD)
+
+##### 2. Proof Key for Conde Exchange [PKCE]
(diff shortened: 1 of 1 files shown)


design/client_conformance_profiles.md
```diff
@@ -0,0 +1,550 @@
...
+
+Client's Request Check
+
+* Operations :
+    - handle authorization request
+
+##### 1. Proof Key for Conde Exchange [PKCE]
+The authorization request includes `code_challenge` and `code_challenge_method`.
+
+`code_challenge_method` is `S256`.
```
mposolda: We already have the switch on the client to enforce S256. This is the case for some other things described here.

In general, question is, what to do with the "small" switches, which we already have and how to handle if their value conflict with the switch for the conformance profile? Few options:
1) Remove all the "small" switches and just replace them with the new FAPI switches. However this may be the issue if some clients want more fine-grained settings. For example some clients may want to enforce PKCE as S256, but still want to authenticate with simple "client_secret" method. Which means that switch this client entirely to FAPI profile will be too restrictive for that client. Due to this, it seems that removing the "small" switches is not an option?

2) If some client will be in the admin console updated to use FAPI profile, then we will automatically update all the "small" switches. For example if client is updating to use FAPI read-only profile, then we will automatically update the "PKCE enforce" switch to S256. But it looks that  we can't do such automatic update for everything? For example if admin wants to update client to use FAPI read-write profile, but client uses "client_secret" method, then automatically changing client authentication method won't work IMO. I think we may need to throw the validation error like "Cannot change client to use FAPI read-write. Please update the client authentication method before" etc.

Those are the implementation details, but probably worth to discuss before we start implementing this?

design/client_conformance_profiles.md
```diff
@@ -0,0 +1,550 @@
...
+* Reference:
+    - Authz Server / 5, 6
+    - Public Client / 1
+    - Confidential Client / 1
+
+Error Response : 400 Bad Request (the detail is TBD)
+
+## TODO
+
+### "acr" (Authentication Context Class Reference) treatment
```
mposolda: We hope to start on step-up authentication later this year. But as usual, no promise for anything :) Help from the community with the feedback and implementation will be more than welcome!

tnorimat: @mposolda I've recognized that this "acr" matter has been opened an discussed in [KEYCLOAK-3314 of JIRA ticket](https://issues.redhat.com/browse/KEYCLOAK-3314) and not yet resolved.

I've looked into release-v3.0.3 of the [FAPI conformance test](https://gitlab.com/openid/conformance-suite) and found that it seems not test "acr" values. Therefore, I've left "acr" matter TODO at this time.

However, I hope I would like to resolve this "acr" matter along with your step-up authentication.

If this "acr" matter need to be discussed, [its JIRA ticket's comment](https://issues.redhat.com/browse/KEYCLOAK-3314) is appropriate?

design/client_conformance_profiles.md
```diff
@@ -0,0 +1,550 @@
...
+
+FAPI security profile includes "acr" related items. However, keycloak has not yet determined what and when "acr" value is set.
+It has already been discussed in [KEYCLOAK-3314](https://issues.redhat.com/browse/KEYCLOAK-3314), but still open.
+
+### track update of FAPI Security Profile
```
pedroigor: @tnorimat Shall we design the SPI (I'm favor of having a specific SPI for this) so that we can support different versions of the profiles?

design/client_conformance_profiles.md
```diff
@@ -0,0 +1,550 @@
...
+* **Status**: Draft #1
+* **JIRA**: [KEYCLOAK-11612](https://issues.jboss.org/browse/KEYCLOAK-11612)
+
+## Motivation
+
+To protect APIs that require high level security like providing financial services, there are several API security profiles like Financial-grade API Security Profile.
+
+To enforce the client operating in complying with such the security profile, the client conformance profiles are introduced. 
+
+## Overview
```
pedroigor: I would also add more information about how to enforce profiles for resource servers. I mean, when the client in Keycloak is actually serving protected resources.

As an example, FAPI Read and Write defines some holder-of-keys mechanisms to prove that a token is actually bound to a caller when accessing a resource server. If we have plans to support that OOTB in our adapters, we would need to somehow configure the adapter properly to also conform with the specs.

I can understand if you think this is out of scope and we should only worry about the policy enforcement on Keycloak side. But something to consider when delivering a E2E deployment that is compliant with a profile.

design/client_conformance_profiles.md
```diff
@@ -0,0 +1,550 @@
...
+    - REST API directly
+    - Import
+ 
+The following 2 ways of client registrations can only determine and set the client conformance profile to the client.
+
+* Dynamic Client Registration Endpoint
+    - OpenID Connect Client Metadata Description
+        * Use Initial Access Token
+
+When creating the initial access token on the admin console, keycloak provides the pull-down menu for specifying the client conformance profile. The selected client conformance profile is included onto the initial access token to convey it to dynamic client registration endpoint afterwards.
```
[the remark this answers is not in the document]
tnorimat: I'm not sure whether clients that are enforced different FAPI security profile coexist in the same realm. However, IMO, it might be probable to do so. 

It might be probable to create the realm per bank and accommodate the client that only access its Account Service API and the one that also access its Payment Service API. In that case, the former be enforced FAPI Read Only Security Profile while the latter be enforced FAPI Read and Write Security Profile.

Meanwhile, considering the point that complete Client Conformance Profiles implementation needs a lot of code change and effort, it seems to be reasonable to implement this Client Conformance Profiles step by step.

Therefore, how about implementing client registration check per realm by using Client Registration Policy as you said at first and implementing client registration check per client later as follow-up task?

BTW, this design document states that implementing client registration check can be conducted on the following two points.

* Dynamic Client Registration Endpoint
    - OpenID Connect Client Metadata Description
        * Use Initial Access Token

* Admin Console
    - UI

Client Registration Policy covers the former. As for the latter, We need to do the same check as Client Registration Policy. Therefore, how about implementing client registration check on the Dynamic Client Registration Endpoint at first as follows

* Dynamic Client Registration Endpoint
    - OpenID Connect Client Metadata Description
        * Use Initial Access Token
        * Anonymous

 and implementing client registration check on Admin Console later as follow-up task?

design/client_conformance_profiles.md
tnorimat: As for 1, IMO, we need to keep "PKCE enforce" switch. PKCE is not a part of FAPI Security Profile Specification. Some user wants to use "PKCE enforce" feature standalone.

As for 2, I think that it cost a lot for us to implement auto-updating feature about client settings. 

Therefore, as you said, it might be better to implement validation and throwing error when the admin updates FAPI Security Profile setting switch on Admin Console.

At this time, if the admin wants to change FAPI Security Profile setting, 

* He needs to do in advance appropriate setting for each client's setting items in comply with FAPI Security Profile setting to be enabled.
* After that, he changes the FAPI Security Profile setting.

It seems not user-friendly but be acceptable on the first implementation of this Client Conformance Profile.

Of course, it might be better to do auto-updating each client's setting items in comply with FAPI Security Profile setting to be enabled. How about realizing it as follow-up task later?

As for 3, Yes, I agree with you.

BTW, considering other FAPI based security profile support in the future (e.g. UK's OpenBanking Security Profile, Australia's Consumer Data Rights(CDR)), it might be better that we add Security Profile pull-down menu (FAPI Read Only, FAPI Read and Write, ...) instead of FAPI switch.

Summarizing up, my proposal is as follows :

* [PKCE Enforce Setting] Keep "PKCE enforce" switch of Client Settings on Admin Console.

* [Security Profile Change] When the admin changes the client's Security Profile setting (e.g. none -> FAPI RW), keycloak checks whether the current client's settings satisfy the Security Profile to be enabled (e.g. FAPI RW). If not, keycloak shows an appropriate message in order to tell the admin what he should do.

* [Client Setting Change under Security Profile Enabled] When the admin changes some client's settings that does not satisfy the current enabled Security Profile setting, keycloak shows an appropriate message in order to tell the admin what's wrong.

WDYT?

design/client_conformance_profiles.md
tnorimat: @pedroigor Yes, I agree with you. We prepare new SPI for this Client Conformance Profiles. 

My proposals is as follows :

* Implement the providers for the latest version (currently, implementer's draft 2 for both FAPI RO and FAPI RW).
* Additionally, implement the providers for the FINAL version when released and replace the previous providers with them.
* keycloak's UI does not show the version (e.g. FAPI RW Security Profile Implementer's Draft 2, FAPI RW Security Profile Implementer's Draft 3, FAPI RW Security Profile Final Draft), only show the name of Security Profile (e.g. FAPI RW Security Profile) due to the following reasons:
  - The slight differences will be expected between FAPI Security Profile's versions after current implementer's draft 2.
  - The user might not understand those slight differences between FAPI Security Profile's versions. If the version would be shown in UI, they might be confused.

WDYT?

design/client_conformance_profiles.md
tnorimat: I've recognized that the client and the resource server also need to operate in comply with FAPI Security Profile. FAPI Security Profile specification describes requirements against the client and resource server.

At this time, I planned to only treat the authorization server, namely keycloak. Therefore, my proposal is as follows :

* [Scope] It is explicitly stated that this design document only treats the authorization server, namely keycloak.

* [FAPI Security Profile Support for keycloak client adapter] If there is a demand for FAPI Security Profile Support for keycloak client adapter and need to be supported, it might be better that the separate design document for FAPI Security Profile Support for keycloak client adapter  be newly created.

WDYT?

design/client_conformance_profiles.md
pedroigor: If companies are stuck with a specific version due to regulatory compliance (e.g.: regulatory framework not yet reflecting latest specs) I would say we should give them an option to choose a version.

You are probably in a better position than me to say whether or not this is a valid concern.

design/client_conformance_profiles.md
pedroigor: Looks good. We should make clear the scope so that people take care of their deployments accordingly in order to adhere to the server policies / requirements.

design/client_conformance_profiles.md
pedroigor: I'm not sure if automatic updating settings is a good idea:

* For some settings we would need to guess the option we find best so that we are assuming risks by choosing one.
* Like stated before, some settings have a huge impact o how the client interacts with the server as well as with clients
* FAPI is basically a set of recommendations and obligations in order to provide a "bullet-proof" OpenID deployment. As such, I think we should recommend settings and delegate the actual work to admins (related to first bullet above)

In regards to recommendations, in addition to the policy enforcement shall we also provide a good UI so that we list what a client is missing in order to comply with FAPI?

design/client_conformance_profiles.md
tnorimat: I see, I agree with you. So what about the following?

* when the admin changes the client conformance profile but some settings violates against the new profile, show the separate UI to show the recommended settings per profile.

design/client_conformance_profiles.md
tnorimat: OK. I'll revise the design document to state these scopes explicitly. Thanks.

design/client_conformance_profiles.md
tnorimat: Yes, I see as you said. So what about the following?

* keycloak's UI shows the profile including its version (e.g. FAPI RW Security Profile Implementer's Draft 2, FAPI RW Security Profile Implementer's Draft 3, FAPI RW Security Profile Final Draft) so that the admin can adhere to the specific version of the profile.

design/client_conformance_profiles.md
tnorimat: I see, I agree with you. So what about the following?

* when the admin changes the client conformance profile but some settings violates against the new profile, show the separate UI to show the recommended settings per profile.

tnorimat:
@stianst

I see your point.

There are 3 points below I've concerned.

* You referred ClientPolicyProvider. It seems to be for Authorization Service. I'm afraid that it cannot be used for this document's purpose(client policy instead of client conformance profiles).

* I think you said that we newly introduce Client Policy(e.g. Client Policy tab on Realm Settings), Condition, Policy Provider instead of Client Conformance Profiles support. Is that right?

* If so, should I close the PR of this Client Conformance Profiles suport desing document and write the design document based on the new concept you said here?

stianst:
ClientPolicyProvider would have nothing to do with authorization services. Guess the naming is a bit confusing. It is rather related to the Client Registration policies we already have.

In essence I believe it would be better to do one thing for:

1. Setting policies on what configuration a client can have
2. Validation of client configurations
3. Conformance to a profile such as FAPI

Today, we have some support for 1, but it's not very flexible and only used for dynamic client registration, but should apply to clients registered through any means available.

We recently had some support for 2 added, but it's not very flexible and very limited.

So instead of introducing a third concept, we can rather merge everything into one thing :)

We could perhaps keep this alive for now, then start a new alternative approach in a different proposal following what I'm suggesting? If you would like to work on it that'd be great, if not I can write something up, but would probably take a little while until I could get around to it.

tnorimat:
OK, I'll encompass these 3 points you've mentioned and re-consider the new design.

As for how to proceed this work, I'd like to continue discussion on this matter in this PR. After clarifying this new design to some extent, I'll write the new design document based on the new approach you've suggested and send PR separately. What do you think about it?

stianst:
Sounds good

tnorimat:
@stianst 

Could we support both the new policy we consider (called Client Policy) and the current Client Registration Policy, or support only the Client Policy covering the current Client Registration Policy and remove the current Client Registration Policy? I think that the latter is preferable in the end.

My proposal is as follows :

1. Implement the basis of the Client Policy. Not yet implement actual Conditions and Executions. At this time, both the Client Registration Policy and Client Policy co-exist in keycloak, but the latter be deactivated.

2. After implementing actual Conditions and Executions that do the same things by the current Client Registration Policy, remove the Client Registration Policy and activate the Client Policy.

WDYT?

VinodAnandan:
@tnorimat @stianst 
OpenID Foundations (@tlodderstedt , @sakimura ) has a proposal to rename and adjust FAPI profiles. More information can be found on this link 

https://bitbucket.org/openid/fapi/issues/271/rename-and-adjust-fapi-profiles

tnorimat:
@VinodAnandan Thank you for notification. I'll follow [it](https://bitbucket.org/openid/fapi/issues/271/rename-and-adjust-fapi-profiles).

PR #54, opened 2020-03-23
KEYCLOAK-10319 Long-term Keycloak Operator architecture

slaskawi:
This Pull Request contains a High Level Architecture overview of Keycloak Operator and its long-term vision. 

Better preview is available here: https://github.com/slaskawi/keycloak-community/blob/Operator-high-level-architecture/design/operator-architecture.asciidoc


design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+We envision Keycloak Operator as:
+
+* an opinionated way of installing and maintaining Keycloak installation on both Kubernetes and OpenShift.
+* an example of a recommended way of installing, configuring, managing custom extension and themes.
+* the thinnest possible layer on top of Keycloak Container and Keycloak Server.
+
+## Introduction
+
+Keycloak Operator was re-designed from the scratch by a small team withing RHMI and RHSSO groups. The implementation plan was described in the [Initial Operator Design Document](operator.md). At the end of a 3 sprint cycle, the team delivered a Level 4 Operator, that was pushed into the [Operatorhub](https://operatorhub.io/operator/keycloak-operator). Once the initial implementation was done, the Operator was handed over to the Keycloak/RHSSO Team.
```
davidffrench: The [Initial Operator Design Document](operator.md) and [Operatorhub](https://operatorhub.io/operator/keycloak-operator) links are not working here. I don't believe they are formatted for asciidoc correctly

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+After maintaining the operator for several months, it became necessary to create a High Level Design (and a long-term vision) how all Keycloak components work together.
+
+## Layered architecture
+
+Keycloak ecosystem consists of 3 layers:
+
+* https://github.com/keycloak/keycloak[Keycloak Server]
+* https://github.com/keycloak/keycloak-containers/tree/master/server[Keycloak Container Image]
+* https://github.com/keycloak/keycloak-operator[Keycloak Operator]
```
davidffrench: I would reverse the order of these to line up with the diagram

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+Implementing it in the Server doesn't make much sense as it will require manipulating `standalone.xml` file based on some environment property.
+
+It can also be implemented in the container as well as in the Operator. However, we push it down the stack, so the correct place for the implementation is the container.
+****
+
+## Components created by the Operator
+
+The diagram below presents the most important connections between components:
+
+image::img/Keycloak-operator-components.png[Keycloak Operator components]
```
davidffrench: Do you think it is worth listing the monitoring resources created as well? A GrafanaDashboard, PrometheusRule, ServiceMonitor and PodMonitor? all if you have the Prometheus CRDs on the cluster.

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+When Keycloak Operator spins up Keycloak installation, it creates:
+
+* `keycloak-db-secret` - used to store database username, password and other properties, such as external address (if used).
+* `credentials-<<CR Name>>` - Admin user credentials to log into Keycloak installation
+* `keycloak` StatefulSet with HA support
+* `keycloak-postgresql` - Responsible for spinning up Postgresql deployment
+* `keycloak-discovery` Service - used for `JDBC_PING` discovery
+* `keycloak` Service - Used for connecting Keycloak using HTTPS (HTTP is not allowed)
+* `keycloak-postgresql` - A service for connecting both internal and external (if used) database instance
+* `keycloak` Route or Ingress - Used for accessing Keycloak
```
davidffrench: Might be worth listing that one is for an OpenShift install and the other for native Kubernetes

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+## Implementation guideline
+
+* **Supporting both OpenShift and Kubernetes is a must** - every feature must work in both usecases correctly.
+* **We prefer convention over configuration** - we hardcode secret names, so that they can be easily figured out by automated scripts.
+* **Do not use mocks** - we just hate them... and with our approach to the architecture, they are completely unnecessary.
+* **We prefer unit tests over e2e tests** - although we do need both, we prefer testing features on the lowest possible layer.
+
+## Further reading
+
+* [Operator Best Practices](https://github.com/operator-framework/community-operators/blob/master/docs/best-practices.md)
```
davidffrench: Same as above, link not working here.

stianst:
Why .asciidoc when all other designs are in markdown? Would be good to just keep it all in one format

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+* https://github.com/keycloak/keycloak-containers/tree/master/server[Keycloak Container Image]
+* https://github.com/keycloak/keycloak-operator[Keycloak Operator]
+
+image::img/Keycloak-operator-layers.png[Keycloak ecosystem layers]
+
+Keycloak Operator uses Keycloak Container image, which in turn uses Keycloak Server bits. The goal is to implement features in the lowest possible layer, so that they get inherited by the all upper layers.
+
+.Feature implementation example
+****
+Let's consider a simple feature - adding JSON formatted logs based on an environmental variable.
```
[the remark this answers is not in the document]
mposolda: TBH When I go through this example, I don't have much clue why the "correct place for the implementation is the container." .

First when looking at the feature description "adding JSON formatted logs based on an environmental variable." - it is not clear to me what exactly this means and which logs are we talking about? Is it possible to add some more context about this?

Then in another sentence, when mentioned this "Implementing it in the Server doesn’t make much sense as it will require manipulating standalone.xml file based on some environment property." I don't understand this TBH...

Just a note that I am noob, who don't have any practical experience with the operators in general. But usually when I want to understand some technology, the examples/quickstarts is the first place, which I am looking at. And from this example I am not able to understand much :) 

Is it possible to add some more context? Or eventually add some links with the expected knowledge for users who look at this design what they need to understand first?

Just my 2 cents when looking at this design from the operator noob perspective :)

design/operator-architecture.asciidoc
slaskawi: Thanks for the comment. Fixed.

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+## Manifesto
+
+We envision Keycloak Operator as:
+
+* an opinionated way of installing and maintaining Keycloak installation on both Kubernetes and OpenShift.
+* an example of a recommended way of installing, configuring, managing custom extension and themes.
+* the thinnest possible layer on top of Keycloak Container and Keycloak Server.
+
+## Introduction
```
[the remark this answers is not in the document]
slaskawi: Let me rename this paragraph to "history".

design/operator-architecture.asciidoc
slaskawi: Fixed. Thanks!

design/operator-architecture.asciidoc
slaskawi: Good point. Fixed.

design/operator-architecture.asciidoc
slaskawi: Good point, let me add it.

design/operator-architecture.asciidoc
slaskawi: Thanks for spotting this. Fixed.

design/operator-architecture.asciidoc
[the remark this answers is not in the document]
slaskawi: Ok, I rephrased it. Please give it another look.

slaskawi:
> Why .asciidoc when all other designs are in markdown? Would be good to just keep it all in one format

@stianst I wanted to use a block with a feature implementation example. This functionality is not available in Markdown. To me, Asciidoc formats documents much better and we also use it in other projects, such as Keycloak Web and Keycloak Documentation.

slaskawi:
> Personas and use-cases - describe related personas and use-cases for the personas

@stianst Added usecases

> More details on being opinionated and providing a high-level usability, rather than low-level toggles

@stianst added "a way to deliver a high level usability rather than low-level toggles" to the manifesto.

> How we will avoid feature creep - I can easily see lots of request for small things to be added, which in the end results in a complicated beast rather than a simple to use opinionated

@stianst To me, this sounds exactly the (although in different words) than the previous point.

>Avoiding logic in the operator - a bit more expanding on the section on where features should be implemented to describe that advanced logic doesn't belong in the operator

@stianst Added "the thinnest possible layer, **with the least possible logic**, on top of Keycloak Container and Keycloak Server." to the manifesto.

> Section capability levels and how we plan to properly address them

@stianst I can add you a dedicated section, but let me push the use cases first. Maybe that will be enough for you.

slaskawi:
@davidffrench @stianst @mposolda Thank you gentlemen for the review. I addressed your comments and pushed a new version.

pedroigor:
@slaskawi Does make sense to list of CRDs used by the Operator? I found one reference to the `Keycloak` CRD but not others currently supported by the operator.

Considering they play an important role, maybe a specific section about CRDs could make this part of the architecture more clear. Especially if provided:

* Their lifecycle. For instance, those implicitly managed by the operator such as the `Keycloak` CRD and the others representing the state that map to entities in Keycloak such as `User`, `Realm` `Client`, etc.

* Limitations and Use-Cases. Is it feasible to manage 100 realms + 1000 users + 100 clients using the operator. Considering Openshift, how that would work when the user is managing the cluster through the Openshift Console ? Is the operator going to support all the configuration we have in Keycloak for these entities?

* Reconciliation. How reconciliation works in regards to CRDs. What happens when updating a CRD or the corresponding entity in Keycloak? Would the operator keep both sides in sync all the time ?

* Relationships between CRDs. How the operator knows that a `User Jown` CRD is mapped to a `Realm My Realm`?

slaskawi:
> Does make sense to list the CRDs used by the Operator? I found one quick reference to the Keycloak CRD but not others currently supported by the operator.

@pedroigor This is coming in updated documentation, see https://github.com/keycloak/keycloak-documentation/pull/796

> Their lifecycle. For instance, those implicitly managed by the operator such as the Keycloak CRD and the others representing the state that map to entities in Keycloak such as User, Realm Client, etc.

@pedroigor I don't want to be too specific about those CRs in this architecture vision as I see them as an implementation detail. However, this is the second time I see someone asking for a list of CRs, so let me add it.

> Limitations and Use-Cases. Is it feasible to manage 100 realms + 1000 users + 100 clients using the operator. Considering Openshift, how that would work when the user is managing the cluster through the Openshift Console ? Is the operator going to support all the configuration we have in Keycloak for these entities?

@pedroigor I see those are Keycloak limitations (not necessarily the Operator). Perhaps we should have a Keycloak long term vision somewhere and put this there? What do you think, @stianst ?

> Reconciliation. How reconciliation works in regards to CRDs. What happens when updating a CRD or the corresponding entity in Keycloak? Would the operator keep both sides in sync all the time ?

@pedroigor I believe this has been covered in Operator SDK manual, so repeating it here doesn't make much sense (to me).

> Relationship between CRDs. How the operator knows that a User Jown CRD is mapped to a Realm My Realm CRD?

@pedroigor That's a good point. Let me mention here the way it has been implemented.

slaskawi:
@pedroigor Pushed new version.

davidffrench:
This looks good to me @slaskawi

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+* https://github.com/keycloak/keycloak-containers/tree/master/server[Keycloak Container Image]
+* https://github.com/keycloak/keycloak[Keycloak Server]
+
+image::img/Keycloak-operator-layers.png[Keycloak ecosystem layers]
+
+Keycloak Operator uses Keycloak Container image, which in turn uses Keycloak Server bits. The goal is to implement features in the lowest possible layer, so that they get inherited by the all upper layers.
+
+.Feature implementation example
+****
+Let's consider a simple feature - adding JSON formatted logs based on an environmental variable.
```
mposolda: Thanks for the updates @slaskawi . But TBH from the description of the feature "adding JSON formatted logs based on an environmental variable." I am personally still not able to see what is the purpose of this feature, which logs are we talking about etc.

It is maybe clear if you have some previous context and knowledge, however for the no-ob like me, it is not clear at all :-) Either adding some more context may be nice or add some "prerequisites" what is the expected knowledge for the people, who are reading this design document (EG. Some Openshift operator basics knowledge or something else).

slaskawi: This is a made up feature just to present the idea of pushing features down the stack. This is definitely not a real feature.

If this answers your comment, please resolve it to make this design more readable.

mposolda: Ok, Thanks

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
+# Keycloak Operator High Level Architecture
+
+* **Status**: Draft #1
+* **JIRA**: https://issues.redhat.com/browse/KEYCLOAK-10319[KEYCLOAK-10319]
+* **Source of the images**: https://docs.google.com/presentation/d/13N5ClXXXcxKjXgor72rUdtMBN07_GneUkhTTwBsy_tE/edit?usp=sharing[Google Slides]
```
stianst, line 5: Remove this link as it is not public

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+
+This document shows a big picture how all the components and layers of Keycloak Operator work together. The blueprint presented here shall be used a guideline for designing features as well as integrating community Pull Requests.
+
+## Manifesto
+
+We envision Keycloak Operator as:
+
+* an opinionated way of installing and maintaining Keycloak installation on both Kubernetes and OpenShift.
+* a way to deliver a high level usability rather than low-level toggles
+* an example of a recommended way of installing, configuring, managing custom extension and themes.
```
stianst, line 15: Not sure I understand this line.

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+## Use cases
+
+Our vision for Keycloak Operator is driven by use cases. The use cases mention the following personas:
+
+* Cluster Admin - a Kubernetes/OpenShift administrator
+* Service Infrastructure Admin - an administrator who provisions Keycloak Operator as well as Keycloak Cluster
+* Realm Admin - a Keycloak administrator, who can provision realms (both using Custom Resources and Admin UI)
+* Secured Application - either configured by an engineering team (using self registration feature) or by Keycloak Admin (or Keycloak Realm Admin).
+
+The use cases we defined are based on Operatr Capability Model:
```
stianst, line 27: Operator, not Operatr

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+* the thinnest possible layer, with the least possible logic, on top of Keycloak Container and Keycloak Server.
+
+## Use cases
+
+Our vision for Keycloak Operator is driven by use cases. The use cases mention the following personas:
+
+* Cluster Admin - a Kubernetes/OpenShift administrator
+* Service Infrastructure Admin - an administrator who provisions Keycloak Operator as well as Keycloak Cluster
+* Realm Admin - a Keycloak administrator, who can provision realms (both using Custom Resources and Admin UI)
+* Secured Application - either configured by an engineering team (using self registration feature) or by Keycloak Admin (or Keycloak Realm Admin).
```
stianst, line 25: I'm not fully convinced about these personas. Looking at the doc I sent you there are:

- Cluster Admin
- Service Infrastructure Admin
- Service Admin
- Service Consumer
- Service End User

I think those are better personas then those you've created. From my perspective the description in context of the Keycloak Operator would be something like:

- Cluster Admin - Provisions cluster resources. Basically, installs the Keycloak Operator.
- Service Infrastructure Admin - Provisions the managed services. Basically, creates the Keycloak CR.
- Service Admin - Manages the Realm. Basically, create the KeycloakRealm CR, but can also manage it through the Keycloak admin console
- Service Consumer - Application developers. Basically, creates the KeycloakClient CR, but can also create clients through other means (admin console, client registration cli, etc.)
- Service End User - Application users. Not sure there are any use-cases for these in this context, but worth mentioning anyways

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+|====
+|Persona |Capability |Use case| Addressed?
+
+|Cluster Admin |Basic Install | Define Pod Disruption Budget to upgrade Kubernetes cluster| Yes
+|Cluster Admin |Basic Install | Use Keycloak as a default IdP in Kubernetes| Yes
+|Cluster Admin |Basic Install | Use Keycloak as a default IdP in OpenShift| No
+|Cluster Admin |Full Lifecycle| Configure Storage Class for Local Backups| Yes
+
+|Service Infrastructure Admin |Basic Install | Install Keycloak Operator | Yes
+|Service Infrastructure Admin |Basic Install | Install and manage Keycloak installation | Yes
```
stianst, line 43: Install and manage Keycloak installation seems a bit broad. What does it cover?

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+
+The use cases we defined are based on Operatr Capability Model:
+
+image:https://raw.githubusercontent.com/operator-framework/operator-sdk/master/doc/images/operator-capability-level.png[Operator Capability Model]
+
+.Keycloak Operator use cases
+[caption=]
+[options="header",cols="^.^,^.^,^.^,^.^"]
+|====
+|Persona |Capability |Use case| Addressed?
```
stianst, line 35: `Addressed` perhaps change this to `Supported` or `Implemented`? Not sure if that's what you mean by Addressed.

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+* **JIRA**: https://issues.redhat.com/browse/KEYCLOAK-10319[KEYCLOAK-10319]
+* **Source of the images**: https://docs.google.com/presentation/d/13N5ClXXXcxKjXgor72rUdtMBN07_GneUkhTTwBsy_tE/edit?usp=sharing[Google Slides]
+
+This document shows a big picture how all the components and layers of Keycloak Operator work together. The blueprint presented here shall be used a guideline for designing features as well as integrating community Pull Requests.
+
+## Manifesto
+
+We envision Keycloak Operator as:
+
+* an opinionated way of installing and maintaining Keycloak installation on both Kubernetes and OpenShift.
```
maleck13, line 13: should upgrades be called out or is that covered under maintaining?

maleck13: nvm I see it is covered

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+* Cluster Admin - a Kubernetes/OpenShift administrator
+* Service Infrastructure Admin - an administrator who provisions Keycloak Operator as well as Keycloak Cluster
+* Realm Admin - a Keycloak administrator, who can provision realms (both using Custom Resources and Admin UI)
+* Secured Application - either configured by an engineering team (using self registration feature) or by Keycloak Admin (or Keycloak Realm Admin).
+
+The use cases we defined are based on Operatr Capability Model:
+
+image:https://raw.githubusercontent.com/operator-framework/operator-sdk/master/doc/images/operator-capability-level.png[Operator Capability Model]
+
+.Keycloak Operator use cases
```
stianst, line 31: I would like to agree on personas before going through the use-cases in more details.

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+|Realm Admin |Deep Insights | Analyze usage and performance metrics| Partially
+|Realm Admin |Autopilot | Scale Keycloak automatically depending on traffic| No
+
+|Secured Application |Basic Install | Accessing Keycloak using a secured connection| Yes
+|Secured Application |Seamless upgrades | Accessing Keycloak during the upgrade process| No
+|Secured Application |Full Lifecycle | Keep session secured even when facing a disaster| Partially
+|Secured Application |Autopilot | Provide stable response time despite the load| No
+|====
+
+## History
```
stianst, line 69: We should just remove the History section - this is irrelevant information.

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+
+.Feature implementation example
+****
+Let's consider a simple feature - adding JSON formatted logs based on an environmental variable.
+
+This feature requires touching all three layers:
+
+1. Keycloak Server needs to provide a JSON log based formatter.
+2. Keycloak Container needs to expose an environment variable, that could be switching on to use JSON formatting
+3. Keycloak Operator needs to expose proper configuration on Custom Resource level enabling JSON log formatting.
```
stianst, line 95: Shouldn't the Operator actually be opinionated here and just use JSON logging since that is the recommended on OpenShift/Kubernetes to have a common/parsable logging format?

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+The basic rule here is to push individual features as low in the stack as possible. In this example, providing JSON log extension only in the Operator would be a mistake.
+****
+
+## Components created by the Operator
+
+The diagram below presents the most important connections between components:
+
+image::img/Keycloak-operator-components.png[Keycloak Operator components]
+
+When Keycloak Operator spins up Keycloak installation, it creates:
```
stianst, line 106: Should make it a bit more clear in the list what types the things below are. For example `keycloak-postgres` is listed twice, once it doesn't say what type it is, then in the second it says in the description it's a service. Then for `keycloak-discovery` it more clearly says it's a Service. I would use the format: `name` `type` - `description` for everything

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+
+* `keycloak-db-secret` - used to store database username, password and other properties, such as external address (if used).
```
stianst, line 108: I do find it a bit weird that the external address for the DB is stored within a secret ;)

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+* `credentials-<<CR Name>>` - Admin user credentials to log into Keycloak installation
```
stianst, line 109: Can you elaborate on this one? I don't understand what this is used for.

stianst:
Please change format to asciidoc to make this consistent with everything else.

lellis32:
hey there, I'm working on a project where I'm considering using the keycloak operator as a component. In my use case, I think it would be useful to be able to add and update Identity/User Storage Providers via CRDs. Would something like that be out of the scope of this project? I mainly just want a way to easily add additional authentication pieces, each bundled with its relevant keycloak configuration, to a big sso system in k8s. And extending this operator to handle those type of CRDs is one idea I had, and doesn't seem terribly hard for me to do, but that could be overkill. Thanks and great job!

design/operator-architecture.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+This document shows a big picture how all the components and layers of Keycloak Operator work together. The blueprint presented here shall be used a guideline for designing features as well as integrating community Pull Requests.
+
+## Manifesto
+
+We envision Keycloak Operator as:
+
+* an opinionated way of installing and maintaining Keycloak installation on both Kubernetes and OpenShift.
+* a way to deliver a high level usability rather than low-level toggles
+* an example of a recommended way of installing, configuring, managing custom extension and themes.
+* the thinnest possible layer, with the least possible logic, on top of Keycloak Container and Keycloak Server.
```
chlunde, line 16: I think I understand what you mean, but this could also be interpreted as just dumping the exported config into map[string]string.

Do you envision being able to set up and configure keycloak just by reading the documentation/CRD and creating a CRD, instead doing the setup in the UI and exporting the configuration (including the "defaults" that are included in the export)?

slaskawi: Our CRDs reflect Keycloak resources. A good example is `RealmRepresentation` JSON payload and `KeycloakRealm` CRD. They are very similar and, in the long-term, should be the same.

So it is perfectly doable to configure Keycloak purely through CRDs (without touching the UI).

PR #55, opened 2020-03-31
KEYCLOAK-12398 Healthcheck endpoint

slaskawi:
https://issues.redhat.com/browse/KEYCLOAK-12398


design/healcheck-endpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+The preferable approach is to hook into Health Endpoint provided by Wildfly (starting from 16) at:
+
+    http://localhost:9990/health
+
+The endpoint is provided according to the [MicroProfile Health](https://microprofile.io/project/eclipse/microprofile-health) specification.
+
+The implementation should take additional parameters into consideration, including:
+
+* `timeout` - the request gets terminated if not performed within a timeout. The implementation needs to take special care of releasing the thread pool (if necessary). Subsequent calls that take longer than expected should not exhaust the thread pools inside Wildfly.
```
stianst, line 26: I don't see any need to be able to set the timeout on a per-request basis, and this should rather just be a server-side configuration.

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+* We can not discover database failures
+* We can not discover if certain Identity Provider is up and running
+* Undertow subsystem is down when proceeding with migration
+
+In order to address this problems, we should provide a parametrized Healthcheck endpoint in the Server and use it in our container image as well as in the Operator. 
+
+## Parametrizing the Healthcheck URL
+
+The preferable approach is to hook into Health Endpoint provided by Wildfly (starting from 16) at:
```
stianst, line 18: Have you checked if it is possible to add query parameters to the WF health endpoint?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,47 @@
...
+    http://localhost:9990/health
+
+The endpoint is provided according to the [MicroProfile Health](https://microprofile.io/project/eclipse/microprofile-health) specification.
+
+The implementation should take additional parameters into consideration, including:
+
+* `timeout` - the request gets terminated if not performed within a timeout. The implementation needs to take special care of releasing the thread pool (if necessary). Subsequent calls that take longer than expected should not exhaust the thread pools inside Wildfly.
+* `fail-on` - a set of conditions under which the probe fails. The initial implementation should cover:
+  * `db` - failing on database failures
+  * `idp-<prefix>` - failing when an IdP is not accessible
```
stianst, line 29: Maybe we should have an option to fail on all idp's?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,47 @@
...
+  * `ldap` - failing is LDAP connection is down
```
stianst, line 30: LDAP is same as IdPs. There may be multiple LDAP instances, as well as other user federation providers. So should probably be "user-federation-<id>"

design/healcheck-endpoint.md
stianst: `<prefix>` doesn't make sense here, should be `<identity provider alias>`

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+A full example might look like this:
+
+    curl http://localhost:9990/health?timeout=2s&fail-on=db,idp-google\
+
+## Limitations for the initial implementation
+
+The ultimate goal is to support all features with `fail-on` flag. However, this is time-consuming and probably not necessary for an initial implementation. The MVP should at least cover:
```
stianst, line 38: I don't see this as a goal as it's to broad I think. I think we can just remove this section.

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,47 @@
...
+
+Deploying both Keycloak and RHSSO on top of Kubernetes requires defining both Readiness Check and Liveness Check (see the [manual](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)). Currently for Keycloak we use a very naive implementation, that calls `/auth/realms/master` URL. This approach has many drawbacks, here are a few examples:
+
+* We can not discover database failures
+* We can not discover if certain Identity Provider is up and running
+* Undertow subsystem is down when proceeding with migration
+
+In order to address this problems, we should provide a parametrized Healthcheck endpoint in the Server and use it in our container image as well as in the Operator. 
+
+## Parametrizing the Healthcheck URL
```
stianst, line 16: I think there's also a case for check, but not fail on. So something like:

`include-checks=db,idp-google&fail-on=db`

stianst:
Did you send this to dev mailing list?

abstractj (commented):
@slaskawi I kind of agree here with Stian about the timeout. IMO that could be a default for now.

Regards the design of the endpoint, would be nice to cover the expected responses, errors and status codes. Also, use the query parameter `fail-on` to everything is a bit confusing.

design/healcheck-endpoint.md
vmuzikar: > Maybe we should have an option to fail on all idp's?

+1
But I'd keep the option to ping a specific IdP as well.

vmuzikar: What would the IdP check do? Just verify if the IdP is online? That doesn't tell us much, does it? Let's say an IdP changes API and login stops working for us. A simple online check won't discover this.

design/healcheck-endpoint.md
slaskawi: In Kubernetes you can put a CPU constrain on a Pod. If you set it to some low value (e.g. `500m`, which is half of a "virtual CPU") and put some traffic through your cluster, the healthcheck API response might take quite a bit of time. If you're a wise cluster admin, and you know you put a CPU constrain on a Pod you may compensate it by increasing the timeout here. 

I'm for leaving it but setting a sensible default if not specified. The Kubernetes probes use `10s` and I think we should use the same value.

design/healcheck-endpoint.md
slaskawi: Over the time, there's one more option I need to investigate - Microprofile Health. It's part of both Wildfly and Quarkus. I'll dive into it shortly.

design/healcheck-endpoint.md
slaskawi: There are two ways to do that:
- Active - we send a request to an external IDP and see if it's up. Not perfect as you mentioned but it's fairly easy to implement.
- Passive - one of the things mentioned in the [Observability Design](https://github.com/keycloak/keycloak-community/pull/177/files#diff-5c0706d1dc103d26e66e94c428afd346R150) is passive tracking of the number of exceptions from a specific external IdP. We can leverage this mechanism here and implement a switch - if we observe 10 exceptions in a certain amount of time, we assume the external IdP has some problems. Otherwise, we assume it's fine.

I'm leaning towards the passive mode. I know it might be less accurate, but it's actually "free" for us.

design/healcheck-endpoint.md
slaskawi: I totally agree. I'll add this.

design/healcheck-endpoint.md
slaskawi: Ok, I'll do that.

design/healcheck-endpoint.md
slaskawi: I think this gap has just been filled by the https://github.com/keycloak/keycloak-community/pull/177

slaskawi:
> @slaskawi I kind of agree here with Stian about the timeout. IMO that could be a default for now.

@abstractj I added some clarification. I'm totally fine with setting a sensible timeout and use it as default. But I think the parameter should stay there because of saturated CPU use cases.

> Regards the design of the endpoint, would be nice to cover the expected responses, errors and status codes. Also, use the query parameter fail-on to everything is a bit confusing.

@abstractj That's a very good point. I'll do that in the next iteration.

design/healcheck-endpoint.md
vmuzikar: +1 for the passive mode, I like it! But we could perhaps combine both approaches.

design/healcheck-endpoint.md
slaskawi: Addressed in the next iteration. Resolving.

design/healcheck-endpoint.md
slaskawi: Perhaps... But I would need very specific examples here.

design/healcheck-endpoint.md
slaskawi: Addressed in the next iteration. Resolving.

design/healcheck-endpoint.md
slaskawi: Addressed in the next iteration. Resolving.

slaskawi:
Thank you all for very good comments. I uploaded the next iteration of the draft.

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+* Undertow subsystem is down when proceeding with migration
+
+In order to address these problems, we need provide a parametrized Healthcheck endpoint in the Server and use it in our
+container image as well as in the Operator.
+
+## Parameterizing the Healthcheck URL
+
+The new Health Endpoint need to support query parameters in the endpoint in the following form:
+
+    http://localhost:8080/auth/health?timeout=2s&fail-on=db,idp-google
```
emmanuelbernard, line 25: why are these choices (fail-on etc) externalized? Shouldn't they be part of a Keycloak deployment choice vs who calls the healthcheck? Isn't the healthcheck caller dump?

slaskawi: Could you please elaborate on this a bit more and give some examples?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+        "cause": "failing for 30s"
+      }
+    }
+  ]
+}
+```
+
+## The workaround for the database upgrade
+
+Since Undertow subsystem starts after the database migration, there's no way provide a working REST Health Endpoint
```
emmanuelbernard, line 133: Can't you just have a database migration service separated from the keycloak other elements? Greenlighting the start?

kenfinnigan:
Was there any thought of bringing the concerns and problems with directly utilizing SmallRye Health back to the project itself: https://github.com/smallrye/smallrye-health?

For one, we've had a goal for the last 6 months to provide pure Java implementations of SmallRye projects with CDI bits layered on top. If SmallRye Metrics is a project that hasn't got there yet, maybe the team could help it to get there to provide a more re-usable implementation?

From an overall perspective, I'd be less concerned with what an MP spec might do and focus on aligning with what a SmallRye project does, especially if it's purely internal to Keycloak and not something end users are creating.

Please bring concerns and problems with the SmallRye Health implementation to the project. Without that feedback it can't be improved and we end up replicating similar code over many projects.

design/healcheck-endpoint.md
kenfinnigan, line 25: Is an example of a URL to be set for Kubernetes Readiness/Liveness probe?

Is it based on having a "generic" health check to feed the parameters into for execution? Is there concern it leaves the URL prone to breakage from incorrect query parameters?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+The implementation needs to take special care of releasing the thread pool (if necessary).
+Subsequent calls that take longer than expected should not exhaust the thread pools inside Wildfly. If not specified,
+`10s` will be used as default (the same timeout as the one used in Kubernetes).
+
+## Implementation
+
+From the implementation perspective, there are two ways of implementing Healcheck calls within Keycloak:
+* Active - whenever we check if a database is down (this is just an example), we initiate a new connection and
+check if it succeeds.
+* Passive - we position the implementation based on Metrics components (see the Observability Design). We actively
```
kenfinnigan, line 43: Can you link to "Observability Design"?

I'd like to read up on what's proposed there

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+
+The former approach seems more natural but has a lof of drawbacks, including:
+* Additional connections allocated in the thread pools.
+* Often requires integrating with additional subsystems (e.g. JCA for database checks).
+* Complicated to ensure that `timeout` parameter is properly respected.
+* Probing External IdPs might be complicated (e.g. what endpoint to call? what about authentication?)
+
+All the above pain points are not applicable for the Passive Mode. Essentially, all database, User Federation LDAP servers
+and External IdPs probes are free of charge as we report success/failure calls for them in the Metrics components.
+With this approach, the Healthcheck Endpoint becomes an alternative reporting tool for the Metrics.
```
kenfinnigan, line 56: Healthcheck endpoints are not designed as a reporting tool for metrics.

They can certainly use metrics to inform their response for liveness/readiness/etc, but it shouldn't be used as a means of reporting metrics that someone is looking at.

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+looks like the following:
+
+    # Liveness Check
+    curl https://<keycloak-url>:8443/auth/health
+
+    # The most basic Readiness Check
+    curl https://<keycloak-url>:8443/auth/health?fail-on=db
+
+    # The more advanced Readiness Check
+    curl https://<keycloak-url>:8443/auth/health?fail-on=db,idp-google,user-federation-myldap
```
kenfinnigan, line 106: Is there a reason it can't be defined as it's own health check group?

design/healcheck-endpoint.md
ebullient: You're reading this backwards, and I think they're correct here.

If they have already reported/recorded success/failure calls for metrics, using those trends to flag health status makes sense (there is a healthcheck registry in micrometer, e.g., for the same reason: to use calculated sla numbers as a health indicator)

design/healcheck-endpoint.md
slaskawi: Let me answer your comment as I understand it...

Keycloak can be deployed in many different scenarios. In one deployment scenario, it might be just a thin layer to let users log in using different External IdP (Google, Facebook, Github etc.). In such case, if the External IdP you rely on is down, Keycloak also can't be considered as "healthy". 

Another example is a full blown SaaS approach. In this scenario you probably use a mix of different External IdPs, User Federation etc. In this example, if any of the External IdPs is down, there's nothing to worry about as user can log in using different methods. 

In the first case, you'd probably want to include the External IdP check in your readiness probe whereas in the latter not.

I believe the probe is very flexible this way.

design/healcheck-endpoint.md
slaskawi: Unfortunately I don't know the history behind the current implementation. Perhaps @stianst @hmlnarik or @mposolda could chime in and give us some more context. I'm not sure if this is achievable or not.

design/healcheck-endpoint.md
slaskawi: > Is an example of a URL to be set for Kubernetes Readiness/Liveness probe?

Yes.

> Is it based on having a "generic" health check to feed the parameters into for execution? Is there concern it leaves the URL prone to breakage from incorrect query parameters?

That's an interesting angle. I believe we'll expose our own Servlet/JAX-RS Endpoint to expose the Healthcheck. In such case, we could emit HTTP 4xx to indicate user error. 

As for the Kubernetes deployments - our suggested way of doing this is the Keycloak Operator. We will probably have some test to ensure the parameters are fine in that case.

design/healcheck-endpoint.md
slaskawi: Of course! It's in the review: https://github.com/keycloak/keycloak-community/pull/177

Just to give you some more context - the Healthcheck is part of the Observability effort. Those two should be considered together.

design/healcheck-endpoint.md
stianst: Actually, this made me think a bit, that it might be better to have some option on the server/realm level to configure what external services should be included in the health check, rather than making it the responsibility of the caller of the endpoint. You may for instance start out with using local username/passwords, but then for whatever reason decide to have a single external IdP instead. In this case Keycloak could actually be smart enough to already know this is an IdP it can't live without, so why then require someone to update the healthcheck URL? Also, how would you update the healthcheck URL when using the operator?

design/healcheck-endpoint.md
slaskawi: I'm sorry, I probably used wrong words here....

The Healthcheck API Design is part of a larger story - the Observability (see [the design](https://github.com/keycloak/keycloak-community/pull/177)). One of the aims of the Observability effort is add more metrics into Keycloak. This Design assumes, that the newly implemented metrics are already implemented. Plus, we have a centralized component, that we can query to get them. 

Having the above in mind, we can query JCA metrics (in case of WF-based deployment) and check the number of failures. If they exceed a certain threshold (to be clarified during the implementation) we could fail the database check. As you can see, both Healthcheck API as well as metrics work in symbiosis in this case. Moreover, the Healthcheck API doesn't do anything actively. It only inspects already collected metrics. So in some sense, from the inner-implementation perspective, it's just another tool that rely on the metrics. 

I hope I made it a bit more clear :)

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+* Passive - we position the implementation based on Metrics components (see the Observability Design). We actively
+analyze database connection metrics reported by JCA subsystem
+(see [the currently available metrics in Keycloak](https://docs.google.com/spreadsheets/d/1YGn4wwUupgQsFTmlm-5VXgLYkigTdy_TEnDhqFxNXXo/edit#gid=0))
+and if we discover  increasing failure count, we report an at Health Endpoint as well.
+
+The former approach seems more natural but has a lof of drawbacks, including:
+* Additional connections allocated in the thread pools.
+* Often requires integrating with additional subsystems (e.g. JCA for database checks).
+* Complicated to ensure that `timeout` parameter is properly respected.
+* Probing External IdPs might be complicated (e.g. what endpoint to call? what about authentication?)
```
stianst, line 52: Actually, for the db check it should be active I'd say, but active part is done by the datasource, where it should keep at least one connection, and have a check on the connection periodically. This would mean that an active Keycloak would just rely on regular db requests, while an idle Keycloak would have a "ping" to the DB.

design/healcheck-endpoint.md
slaskawi: If I'm reading you correctly, you're suggesting aligning with the [Groups proposal for Smallrye Health](https://xstefank.io/2020/04/21/health-groups-experimental-microprofile-health-extension-in-quarkus/). Being more specific, the call would look like this:

    curl https://<keycloak-url>:8443/auth/health/group/db
    curl https://<keycloak-url>:8443/auth/health/group/user-federation-myldap
    curl https://<keycloak-url>:8443/auth/health/group/idp-google

If you'd like to check them all, you'd call:

    curl https://<keycloak-url>:8443/auth/health/group

This is a very interesting idea. Martin told me that the Groups concept should be pushed into Microprofile Health pretty soon. That would mean we do something more or less standardized over the industry. I like this idea.

@stianst @vmuzikar @abstractj @ASzc @hmlnarik @miquelsi Perhaps you'd be interested in sharing your thoughts here?

design/healcheck-endpoint.md
stianst: Problem with this approach is that an idle server does not have health checks. Not sure that is a big problem with user federation or external IdPs though. For external user federation we could effectively use the same strategy as for DB, where a connection pool with at least one connection is kept open. For external IdPs that's harder, especially in cases where there are 100s of external IdPs.

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+or integrating with Microprofile Health Subsystem. An additional problem is that Wildfly provides its own
+`HealthReporter` class (see [the code](https://github.com/wildfly/wildfly/blob/master/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporter.java))
+that doesn't use groups. The Grouping functionality is essential for this design (explained below).
+
+The biggest downside of providing Keycloak's own implementation is that we'll end up with 2 health endpoints -
+one for Wildfly (`http://localhost:9990/health`) and one for Keycloak (`https://localhost:8443/auth/health`). We
+need to be very clear in the documentation that the latter is the proper one.
+
+Having the above in mind, the design assumes integrating the new Healthcheck API with Keycloak's REST interface.
+The implementation will create additional `/health` endpoint that will parse incoming query parameters. Once all
```
stianst, line 89: There should obviously only be one health endpoint, so if we have our own we'd need to also run any relevant underlying container health checks. What's the overhead of doing that?

design/healcheck-endpoint.md
stianst: We're going to need a different strategy for DB upgrades (both before and after we support zero downtime upgrades). Not sure what is generally best practices here on K8s.

stianst (commented):
Missing a clear list of what checks we should have

slaskawi:
> Was there any thought of bringing the concerns and problems with directly utilizing SmallRye Health back to the project itself: https://github.com/smallrye/smallrye-health?

@kenfinnigan What concerns do you mean exactly? If you're referring to the lack of CDI support in Keycloak, that's probably not a SmallRye Health project's problem. Probably the only thing that could be improved is to create some sort of static access to the most critical classes. I created an issue for this here: https://github.com/smallrye/smallrye-health/issues/172

> For one, we've had a goal for the last 6 months to provide pure Java implementations of SmallRye projects with CDI bits layered on top. If SmallRye Metrics is a project that hasn't got there yet, maybe the team could help it to get there to provide a more re-usable implementation?

@kenfinnigan I believe the goal has been achieved. SmallRye Health works great with CDI. But during initial design of Keycloak, CDI has been dropped. I've heard that performance was one of the deciding factors but let me defer to @stianst here.

> From an overall perspective, I'd be less concerned with what an MP spec might do and focus on aligning with what a SmallRye project does, especially if it's purely internal to Keycloak and not something end users are creating.
>
> Please bring concerns and problems with the SmallRye Health implementation to the project. Without that feedback it can't be improved and we end up replicating similar code over many projects.

@kenfinnigan I already did that with the ticket I created. Thanks for pointing this out!

slaskawi:
> Missing a clear list of what checks we should have

@stianst The list has been provided in lines 30-32. We currently have only 3 of them:

```
  * `db` - failing on database failures.
  * `idp-<identity-provider-alias>` - failing when an IdP is not accessible.
  * `user-federation-<id>` - failing is LDAP connection is down.
```

If you'd like to extend this list, please suggest a comment inline.

design/healcheck-endpoint.md
slaskawi: I think putting this into server configuration doesn't buy us anything. Whenever you'd like to change the behavior of the probe, you'd probably need to restart the server or cycle the Pods. 

The only option I find interesting is putting the probe configuration into the server runtime. This probably means we'd need to put the configuration parameters somewhere into the database and provide some UI screen with the options. 

But then, the problem appears - the Admin Console is built around the Realm concept. The Readiness/Liveness probe is something that belongs to the server itself, not a particular Realm. So where to put a configuration screen for it? Also, is it worth the effort? I'd say not for the first iteration of this feature. In the future - perhaps?

design/healcheck-endpoint.md
slaskawi: I believe the key requirement is that Liveness Probe needs to be up and Readiness Probe down. This workaround assumes this behavior. 

But we can change the approach to whenever we want. At the moment, Keycloak Operator uses a ConfigMap with a raw Liveness/Readiness checks (see [the code](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_probes.go)). We can put whatever we want there.

design/healcheck-endpoint.md
slaskawi: Let's consider this as two separate cases - DB and External IdPs.

As for the former, the JCA subsystem exposes a couple of [interesting metrics](), including:
- `wildfly_datasources_pool_in_use_count{data_source="KeycloakDS",microprofile_scope="vendor"}`
- `wildfly_datasources_pool_idle_count{data_source="KeycloakDS",microprofile_scope="vendor"}`
- `wildfly_datasources_pool_max_wait_time{data_source="KeycloakDS",microprofile_scope="vendor"}`
- `wildfly_datasources_pool_timed_out{data_source="KeycloakDS",microprofile_scope="vendor"}`
- `wildfly_datasources_pool_created_count{data_source="KeycloakDS",microprofile_scope="vendor"}`
There are many others. I will require some experimentation but I believe we could come up with a pretty good probe only by analyzing the 5 listed above. Another aspect is that JCA subsystem needs to actively probe the datasource. This can be achieved by configuring `check-valid-connection-sql` and `background-validation` (see the [Wildfly Model manual](https://docs.wildfly.org/19/wildscribe/subsystem/datasources/data-source/ExampleDS/index.html)).

External IdPs can be tracked by analyzing the amount of errors in the past few minutes. Perhaps this "few" should be passed into the probe as a parameter or at least there should be some way to override this. If there's no traffic to an external IdP, there will be no errors, thus the probe will pass.

I can think of one, very tricky scenario that I don't have a good answer to. If an External IdP is down for a very long time, Keycloak will flip between ready and not-ready in the predefined intervals. If the Readiness probe is down, Kubernetes will plug Keycloak out of the Load Balancer. Therefore, it will receive no traffic. Since the Readiness probe checks the number of error in the past few minutes, it will eventually report success. Once this happens, Kubernetes will plug Keycloak back into the Load Balancer and Keycloak will receive some traffic. Then, the number of errors reported from an External IdP goes up again and the Readiness probe fails. 

I'm not sure if the above scenario is problematic or not. It's a bit similar to an active Readiness Check where you send some traffic to an External IdP to see if it works again. But in our case, this ends up in a bad user experience. I'm very interested what do you think about this? One solution that comes into my mind is passive/active check. So whenever you're going back from non-ready to ready, you perform an active check (ping the External IdP public URL for example) to see if it's really up.

design/healcheck-endpoint.md
slaskawi: I believe I answered this here: https://github.com/keycloak/keycloak-community/pull/55#discussion_r496539094

design/healcheck-endpoint.md
slaskawi: Wildfly's Healthcheck Endpoint is focused on the server and deployed applications. It checks the following things (see [the code](https://github.com/wildfly/wildfly/blob/cf2ecf8cd575f425ba5cb254023fd6d44f2aa070/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/ServerReadinessProbes.java)):
- The `server-state` attribute.
- Checks the `read-boot-errors` from Core Service
- Checks if all deployments are fine

If the new Healthcheck endpoint is incorporated into Keycloak's codebase, none of the above checks applies. If anything bad happens during the boot time, Keycloak's JAX-RS endpoints will be down and the probe will naturally fail. 

The only questionable thing is checking the deployed Providers. But I'm not sure how to do it properly. Perhaps we could implement an additional check:

    curl https://<keycloak-url>:8443/auth/health?fail-on=provider-my-provider-id

But I'm not keen of doing that. I've always perceived deploying custom Providers to Keycloak as a configuration thing.

design/healcheck-endpoint.md
stianst: From the design here you are stating that DB checks is passive, meaning it will require traffic to Keycloak to detect DB is down. However, in your answer you linked you are saying that it is basically active, since it is keeping one active connection in the pool. So that means the design doesn't reflect the actual plan.

design/healcheck-endpoint.md
stianst: The problem here is that user federation and identity providers are a per-realm thing, while the probe is now an external thing to configure. Now if a realm admin adds a user federation provider, he would then have to communicate to the KC admin to update the probe. For a realm that has a single IdP at the realm level as I mentioned we can actually detect that and know it is important, without requiring any manual update to the probe.

stianst: Another thing is that we now have user fed and idps at the global server level in the health check, while in reality one realm could be up and running just fine with a user fed provider being down, while another realm may rely on it for its service.

design/healcheck-endpoint.md
stianst: KC start should probably really just fail if a custom provider fails, as we can't expect KC to operate in a healthy way without an expected provider. So we don't need a health check for providers as such.

design/healcheck-endpoint.md
stianst: Not sure to be honest. First and foremost we are creating a health/readiness endpoint for K8s right. So how would you then use the groups?

design/healcheck-endpoint.md
slaskawi: I agree.

design/healcheck-endpoint.md
slaskawi: Yes, perhaps I should clarify this a bit more.

The Passive Mode in this context means that we don't actively force JCA subsystem to check the connections. We just observe the metrics it emits. But on the other hand, the JCA subsystem actively maintains the connection pool and validates connections inside of it.

I'll try to make it a bit more clear in the design document.

design/healcheck-endpoint.md
slaskawi: Groups use different approach to the logic than we use in our design. With our approach it is very easy to include or exclude a certain check (like a database check for example), whereas it's pretty hard to reflect this with Groups.

Let's consider the following checks for example:

    # Readiness Check - checks database and Google External IdP
    https://<keycloak-url>:8443/auth/health?fail-on=db,idp-google
    # Liveness Check - checks only if REST interface is up
    https://<keycloak-url>:8443/auth/health

We could certainly implement this with separate `https://<keycloak-url>:8443/auth/health/ready` and `https://<keycloak-url>:8443/auth/health/live` but let's try to focus on Groups here. The `/health/group` REST interface aggregates all groups. This means it would check the database, External IdPs, LDAP... everything. In our example, we'd like to exclude all External IdP apart from Google as well as all User Federation checks (LDAP etc). The current SmallRye Groups implementation seems to allow calling a single Group per invocation. That means, we'd need to call the probe two times:

     https://<keycloak-url>:8443/auth/health/groups/db
     https://<keycloak-url>:8443/auth/health/groups/idp-google

This is a bit unfortunate, as Kubernetes allows only one HTTP Probe definition (although, we could craft a Bash script up to call both endpoints for us). 

Calling the Liveness example is also problematic, because we want to exclude all checks there. We just want to ping the REST interface and see if its up. The only idea I have in mind is to create a separate Group for it - something like:

    https://<keycloak-url>:8443/auth/health/groups/rest-only

But again, all this seems to be less flexible than the proposal from this Design. 

Let me also hook @xstefank into the discussion. Perhaps he has some neat idea that we could use here.

design/healcheck-endpoint.md
slaskawi: This is a very important question! 

From the user perspective, nothing really changes. Imagine that your deployment requires a Google IdP, which is down for some reason. If you don't include the External IdP check in the Readiness Probe, you will end up with an error screen in Keycloak whenever you try to log in. If you include the External IdP check in the probe, you will get a HTTP 503 Error from Kubernetes/OpenShift as there will be no ready Pods behind a Load Balancer. So the end game is the same - a user can not log in. It's only a matter what Error message he sees.

Also, the same question we should probably ask with User Federation bit. What happens if an LDAP Server is down? Should we ever mark Keycloak as not-ready than?

If we decided to leave both External IdPs and User Federation out of this Design, it will make all the things very simple. The Readiness Check will probably include the database check by default, whereas the Liveness Check will ensure that the REST interface is up. We'd end up with a very simplistic approach that probably won't require any design at all (because it will fit into 2-3 classes).

design/healcheck-endpoint.md
slaskawi: A few comments below (here: https://github.com/keycloak/keycloak-community/pull/55#discussion_r496622654) we started a discussion whether or not External IdPs (and I also added User Federation to this context) belong to the Healthcheck API. Perhaps not? Perhaps they should be only an item in the monitoring dashboard?

design/healcheck-endpoint.md
kenfinnigan: @slaskawi thanks, it does. I was concerned health checks were being treated as more than just utilizing metrics data to determine their status.

design/healcheck-endpoint.md
xstefank: @slaskawi you can only define another group that will combine both `db` and `idp-google` groups into a single URL as mentioned. Health checks can be part of several groups. However, it might be useful to have also exclude options if you have many groups/checks. This is one of the most complex use cases I've seen so far so I would like to understand what is your preferred approach so we can adjust the feature. If you feel like that `fail-on` or `exclude` would be better for you please create an issue in the smallrye-health repo describing the use case.

design/healcheck-endpoint.md
maleck13: @slaskawi  @stianst  I would also fall on the side of a readiness probe purpose is to indicate the service is able to serve requests not a particular subset of requests. I don't think you would want to take your keycloak servers out of the service pool just because an IDP was down?  It would still be perfectly valid for a customer to access the server to configure users, new IDPs etc a 503 in this situation would seem the wrong behaviour. I would suggest a focus on whether keycloak can serve requests overall (IE can I access the keycloak server and do regular keycloak admin) If an IDP is unavailable a metric and informational alert may be enough here?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,159 @@
...
+## Parameterizing the Healthcheck URL
+
+The new Health Endpoint need to support query parameters in the endpoint in the following form:
+
+    http://localhost:8080/auth/health?timeout=2s&fail-on=db,idp-google
+
+The following list of additional parameters will be supported:
+
+* `fail-on` - a set of conditions under which the probe fails. The initial implementation should cover:
+  * `db` - failing on database failures.
```
maleck13, line 30: Is there a situation in which you would not want keycloak to fail its readiness check if the db was down? Should this just be a default behaviour (IE readiness check always fails if DB is inaccessible )

design/healcheck-endpoint.md
slaskawi: After having a discussion with @stianst, we decided to remove both External IdPs and User Federation from the Healthcheck API.

Here's the most important argument for doing this:

> Including both checks probably makes the end-user situation worse. If we include an External IdP check in the Health endpoint, Kubernetes will plug all the pods out of the Service (a load balancer) if that IdP is down. Once all pods are removed from the Service, an end-user will see an HTTP 503 Error page generated by the OpenShift router. If however, the External IdP is down and we don't include it in the Health endpoint, a user will probably face an error during log in. But the error page will be displayed by Keycloak and we control what message we send there. I consider this as a slightly better situation, although the end-result is the same - a user can not log in.

design/healcheck-endpoint.md
slaskawi: Together with @stianst we decided to remove the External IdPs and User Federation from Healthcheck API. See [this comment](https://github.com/keycloak/keycloak-community/pull/55#discussion_r498709614) for more explanation. 

Resolving this conversation.

design/healcheck-endpoint.md
slaskawi: Having in mind that we've decided to remove the External IdPs and User Federation from the Healthcheck API (see [this comment](https://github.com/keycloak/keycloak-community/pull/55#discussion_r498709614)), the database check is the only one that remains. I believe it makes a lot of sense to include it by default. Having in mind, that we perform a Passive checks, this is essentially free of charge for us.

So, yeah, totally agree.

design/healcheck-endpoint.md
slaskawi: Totally agree, @maleck13.

design/healcheck-endpoint.md
slaskawi: I added small clarification about the Readiness Probe.

Let me resolve this conversation than.

design/healcheck-endpoint.md
slaskawi: Let me resolve this conversation. I believe this will be covered in the second draft I plan to sent in a minute.

design/healcheck-endpoint.md
slaskawi: Please do not resolve this conversation. It should stay here as a reference why we decided to leave the External IdPs and User Federation out.

design/healcheck-endpoint.md
slaskawi: Let me resolve this conversation.

design/healcheck-endpoint.md
slaskawi: I believe this conversation can be resolved now. Together with @stianst we decided to remove External IdPs as well as User Federation bits. See [this comment](https://github.com/keycloak/keycloak-community/pull/55#discussion_r498709614).

design/healcheck-endpoint.md
slaskawi: I believe this conversation can be resolved.

design/healcheck-endpoint.md
slaskawi: Added to the new version of the Design. Resolving.

design/healcheck-endpoint.md
slaskawi: Clarified in the new version. Resolving.

design/healcheck-endpoint.md
slaskawi: Thanks you for the input @xstefank. Together with @stianst we decided to simplify the design and leave the idea of the optional checks out.

slaskawi: Let me resolve this conversation.

slaskawi:
Dear all,

I've just uploaded the third iteration of the Healthcheck proposal.

Differences from the previous version:
- I removed both External IdPs and User Federation bits from the scope (you may find the explanation in the design).
- I removed the "fail-on" as there will be checking two things - if the database and the JAX-RS interface is up.
- Liveness probe and Readiness Probe will use different endpoints. We'll use the Liveness Probe from Wildfly until https://issues.redhat.com/browse/WFLY-13931 gets implemented.
- In order to mitigate the bad user experience from the above, we'll use a dedicated "readiness.sh" and "liveness.sh" probes from the Container.

Feedback is more than welcome!

Thanks,
Sebastian

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Healthcheck API for Liveness and Readiness probes
+
+Since Undertow subsystem starts after the database migration, there's no way provide a working REST Health Endpoint
+during the database migration time. The long boot up time may also happen when loading up sessions from the database,
+or a remove Infinispan cluster. In both scenarios, it is necessary to implement a passing Liveness probe, and a failing
+Readiness probe. This will prevent Kubernetes from killing the Pods, but it will also unplug the pods from a Service
+and make them unavailable for the external consumption.
+
+One of the solutions is to use Wildfly's Liveness check available at `http://localhost:9990/health/live`. The check
```
vmuzikar, line 119: Does this mean we'll use WildFly's Liveness check all the time even after KC has started? (That is, at least until WFLY-13931 is resolved.) What if the pod gets into some failing state and needs to be rebooted? WF's endpoint might not be aware of such state. If I understand it correctly, the pod would be just removed from the Service (thanks to the Readiness check) but won't reboot it. Is that enough?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,169 @@
...
+Health endpoint, a user will probably face an error during log in. But the error page will be displayed by Keycloak,
+and we control what message we send there. I consider this as a slightly better situation, although the end-result is
+the same - a user cannot log in.
+- Parametrized checks, e.g. `http://localhost:8080/auth/health?timeout=2s&fail-on=db,idp-google`. After removing
+the External IdPs and User Federation, the parametrized check is no longer necessary. Perhaps, we'll get back
+to it in the future.
+
+## Healthcheck parameters
+
+The new Health Endpoint need to support the `timeout` query parameter in the endpoint in the following form:
```
stianst, line 35: Timeout parameter doesn't make sense to me. What would it actually control?

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,169 @@
...
+and if we discover  increasing failure count, we report an at Health Endpoint as well.
+
+The former approach seems more natural but has a lof of drawbacks, including:
+* Additional connections allocated in the thread pools.
+* Often requires integrating with additional subsystems (e.g. JCA for database checks).
+* Complicated to ensure that `timeout` parameter is properly respected.
+
+All the above pain points are not applicable for the Passive Mode. Essentially, all database, User Federation LDAP servers
+and External IdPs probes are free of charge as we report success/failure calls for them in the Metrics components.
+With this approach, the Healthcheck Endpoint becomes an alternative reporting tool for the Metrics.
```
stianst, line 61: This section doesn't seem to have any relevance

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,169 @@
...
+- `wildfly_datasources_pool_in_use_count{data_source="KeycloakDS",microprofile_scope="vendor"}`
+- `wildfly_datasources_pool_idle_count{data_source="KeycloakDS",microprofile_scope="vendor"}`
+- `wildfly_datasources_pool_max_wait_time{data_source="KeycloakDS",microprofile_scope="vendor"}`
+- `wildfly_datasources_pool_timed_out{data_source="KeycloakDS",microprofile_scope="vendor"}`
+- `wildfly_datasources_pool_created_count{data_source="KeycloakDS",microprofile_scope="vendor"}`
+The implementation should include some experimentation, but the above 5 metrics should be enough to check the proper
+JCA state. The other metrics are available in [this document](https://docs.google.com/spreadsheets/d/1YGn4wwUupgQsFTmlm-5VXgLYkigTdy_TEnDhqFxNXXo/edit?usp=sharing).
+Another aspect is that JCA subsystem needs to actively probe the data source.
+This can be achieved by configuring `check-valid-connection-sql` and `background-validation`
+(see the [Wildfly Model manual](https://docs.wildfly.org/19/wildscribe/subsystem/datasources/data-source/ExampleDS/index.html)).
```
stianst, line 109: This section should cover thoughts on how to do "active" monitoring of database by having a min pool size set to 1, with a background check on connections

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+  ]
+}
+```
+
+Having two Healthcheck endpoints is unfortunate. Once [WFLY-13931](https://issues.redhat.com/browse/WFLY-13931) gets
+implemented, it will be possible to migrate Keycloak's Readiness Check directly into Wildfly. For the time being,
+we a helper `readiness.sh` and `liveness.sh` scripts will be implemented in the Keycloak Container. This
+will enable us to transition between one and another solution transparently.
```
stianst, line 151: I'm not really keen on this approach at all. Is there nothing better we can do?

stianst (commented):
Kube wants a liveness and a readiness probe right? It seems we are only getting a single endpoint, so what is used for what? There's no mention of Quarkus in this design, only WF, so how does this apply to Quarkus (will we be able to do something that is uniform?).

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,169 @@
...
+- Implementation suggestions
+
+After many discussions, we decided to remove from the initial design the following items:
+- External IdP and User Federation checks. Including any of those checks makes the end-user situation worse.
+If we include an External IdP check in the Health endpoint, Kubernetes will plug all the pods out of the Service
+(a load balancer) if that IdP is down. Once all pods are removed from the Service, an end-user will see an HTTP 503
+Error page generated by the OpenShift router. If however, the External IdP is down, and we don't include it in the
+Health endpoint, a user will probably face an error during log in. But the error page will be displayed by Keycloak,
+and we control what message we send there. I consider this as a slightly better situation, although the end-result is
+the same - a user cannot log in.
```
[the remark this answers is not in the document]
slaskawi: For the scope of this design, the environment is Kubernetes or a very similar environment (Liveness/Readiness checks are used by other orchestrators as well). The other thing is that both checks are black or white - they report either a success or a failure and there's nothing in between. This is a place (as you already noticed it), where the Observability comes into play. The SRE Team (a team that looks after Keycloak) will then notice, that the Keycloak cluster is up but an External IdP is down. Depending on multiple factors there might be something to do (e.g. firewall configuration update) or not (if the External IdP is really down).

slaskawi:
Today morning, both @stianst and I had a discussion on the future plans of both Healthcheck API and the Observability designs. 

We agreed that our main focus will be Quarkus-based distribution. As for the time being, we'll use what we have for the Wildfly-based distribution (which translates to the [embedded metrics](https://github.com/keycloak/keycloak-containers/tree/master/server#enable-some-metrics), custom [Aerogear Extension](https://github.com/aerogear/keycloak-metrics-spi) for additional metrics and the [Healthcheck](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_probes.go) as they are now). Both Designs will be repurposed and updated in the following days.

One of the crucial differences is that we agreed to verify if we can use CDI extension in Quarkus (just to provide additional context - the CDI has been removed from Wildfly-based distribution as it introduces additional latency and increased the bootstrap time). Both Microprofile Health as well as Micrometer implementation provide CDI extensions to easily wire everything together. 

//cc @stianst @ASzc @miquelsi @abstractj @vmuzikar @pedroigor

design/healcheck-endpoint.md
slaskawi: Because of https://github.com/keycloak/keycloak-community/pull/55#issuecomment-708356265 there are two situations to analyze here:
- The current Wildfly-based deployment. The answer is yes. If Keycloak gets into such a weird state, it is SRE Team responsibility to look at the SRE Dashboard and verify if Keycloak really hangs. If it does - we need to fix it in our code. The Readiness/Liveness Probe is definitely not a place to do a deep system analysis. It needs to be quick, small and reliable.
- Quarkus-based deployment. In this case Quarkus is tied very closely with Keycloak (it's more a Spring Boot approach rather than JEE Platform approach). Quarkus already provides a built-in Liveness check and we should probably use it. If Keycloak hangs, the procedure is exactly the same as in WF-based deployments.

design/healcheck-endpoint.md
slaskawi: To some degree I agree with you. If we rely solely on passive checks, we might remove it. However, if we ever do any form of active check (e.g. send ping to a database), we to supply some timeout for that operation. 

I guess we can remove it. It should be pretty straightforward to add it if needed.

design/healcheck-endpoint.md
slaskawi: I created this for @kenfinnigan. I wanted to make it crystal clear the implementation will use metrics and reports the status based on them. From the conceptual angle, it becomes a reporting tool. 

Let me leave it as is.

design/healcheck-endpoint.md
slaskawi: I agree. Let me add it.

design/healcheck-endpoint.md
slaskawi: I hope we came to an agreement that the healthcheck complexity should go into the healthcheck code but the scripts are also requires as they enable the SRE Team (or any sort of Ops Team) to implement last-minute fixes and workarounds.

Let's keep the current approach - the Operator creates a ConfigMap with Bash scripts used for both probes. The Probes will invoke curl pointing to the Healthcheck REST. It's super flexible this way.

slaskawi:
Dear all,

After some discussion with Stian, I've created another iteration of the proposal. You may find it here: https://github.com/keycloak/keycloak-community/pull/55

Differences from the previous version:
- The design focuses on Keycloak.x, our Quarkus-based distribution. Wildfly-based distribution is out of the scope.
- Since using CDI is essentially "free" in Quarkus, we allow it's use to integrate with existing Quarkus integration code.
- The design now exposes 2 endpoints: /health/ready and /health/live. Those endpoints are aligned with MP Health.
- I updated the design on database check. We will maintain at least 1, healthy connection to obtain health metrics for passive check.
- The "timeout" parameter has been removed. Although, the implementation will have to use some sort of timeout internally.

In general, this version is very simplistic. I've gone quite a long journey from the initial proposal to what we have now.

Feedback is more than welcome but I highly recommend to comment on the PR directly.

Thanks,
Sebastian

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,132 @@
...
+When Keycloak is serving traffic, the implementation needs to track the database health by observing metrics reported
+by the Agroal framework. However, when Keycloak is idle, the implementation needs to either ask Agroal to maintain at least
+one, healthy connection or perform an active check.
+
+#### Healthcheck API for Liveness and Readiness probes
+
+The REST response has been defined in [Microprofile Health Spec](https://download.eclipse.org/microprofile/microprofile-health-3.0-RC4/microprofile-health-spec.html#_appendix_a_rest_interfaces_specifications).
+There are 3 possible HTTP Status codes that might be returned from the endpoint:
+* 200 - Healthcheck passes
+* 500, 503 - Healthcheck fails
```
hmlnarik, line 97: Looking at the [link from above](https://download.eclipse.org/microprofile/microprofile-health-3.0-RC4/microprofile-health-spec.html#_appendix_a_rest_interfaces_specifications), the 200 status can also provide key-value detailed report.
Could you provide sample response for all of the three status codes? What kind of `data` datapoints would be mandatorily exposed for any health check, if any? Would e.g. `cause` be mandatory for any failed check in 500 / 503 response?

hmlnarik (commented):
Overall LGTM, thanks @slaskawi! I've got only one ask to provide more details, please see inline

stianst (approved):

miquelsi (approved):
Nice work @slaskawi!

design/healcheck-endpoint.md
slaskawi: That's a good point @hmlnarik !

I believe it doesn't make sense to expose anything for a successful check. Successful checks are not logged anywhere in Kubernetes (failed are - in Events).

HTTP 503 is a standard response for a failed probe. It should contain enough debugging information to help users investigating a failure. In our case - this should probably the time that the probe fails. It might also be a good idea to expose an exception message but we need to make sure, there's no secret information with it. 

HTTP 500 indicates an exception thrown in the probe's code.

slaskawi: I slightly updated the text. I hope it works for you.

slaskawi:
I've just pushed a minor review to address @hmlnarik 's comment. 

After the implementation, I'll review this again and make sure all the bits match. After this step - it will be ready to be merged.

hmlnarik (approved):
Looks good to me, nice job, @slaskawi!

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,134 @@
...
+Agroal (the lightweight database connection pool) publishes a set of metrics into the main Quarkus Metrics Registry
+(see [the code](https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/metrics/AgroalMetricsRecorder.java#L37-L113)).
+It also provides an active healthcheck (see [the code](https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java)).
+
+The goal of the implementation is to provide a specialized check that is a hybrid between an active and a passive mode.
+When Keycloak is serving traffic, the implementation needs to track the database health by observing metrics reported
+by the Agroal framework. However, when Keycloak is idle, the implementation needs to either ask Agroal to maintain at least
+one, healthy connection or perform an active check.
+
+#### Healthcheck API for Liveness and Readiness probes
```
alechenninger, line 92: It's not completely clear here, but I don't believe it is best practice for a readiness or liveness endpoint to check external dependencies. What is restarting keycloak pods going to do if the database is down? (Just make a situation worse, would be my guess)

design/healcheck-endpoint.md
```diff
@@ -0,0 +1,134 @@
...
+- Parametrized checks, e.g. `http://localhost:8080/auth/health?timeout=2s&fail-on=db,idp-google`. After removing
+the External IdPs and User Federation, the parametrized check is no longer necessary. Perhaps, we'll get back
+to it in the future.
+
+## Healthcheck endpoint
+
+The new Healthcheck endpoint will contain two checks - Liveness and Readiness:
+
+    http://localhost:8080/auth/health/ready
+    http://localhost:8080/auth/health/live
```
miquelsi, line 42: The Quarkus distribution places Keycloak at the root of the URL, so no `/auth` is needed here.

design/healcheck-endpoint.md
slaskawi: The Healthcheck Endpoint only checks the database at the moment (in it's readiness probe). Perhaps, in the future we will include a remote Infinispan cluster. But this needs a bit of experimentation. 

The Liveness check only checks if the Keycloak REST interface is up. If it's not - something terrible happened at the boot time. So there's a good argument to restart it.

design/healcheck-endpoint.md
slaskawi: Done. Thanks for spotting this!

slaskawi:
@miquelsi @vmuzikar @ASzc @abstractj The last bits have just been added. @abstractj Could you please merge it?

yav124:
Will this still be merged? Would be great to have this health check.

abstractj:
@stianst I completely missed this change, but considering that it was approved long time ago, I would like to double check if we still would like to merge it.

@vmuzikar WDYT?

PR #99, opened 2020-04-23
KEYCLOAK-13933 Client Policies

tnorimat:
This PR is for [KEYCLOAK-13933](https://issues.redhat.com/browse/KEYCLOAK-13933) .

### DIFF
--- a/design/client-policies.md
+++ b/design/client-policies.md
@@ -5,17 +5,20 @@
 
 ## Motivation
 
-To make it easy to secure client applications, it is beneficial to realize the following three points in a unified way.
+To make it easy to secure client applications, it is beneficial to realize the following four points in a unified way.
 
 1. Setting policies on what configuration a client can have
 2. Validation of client configurations
 3. Conformance to a profile such as FAPI
+4. Limiting what type of configuration a client can have based on source that is creating the client
 
 1 and 2 have been realized by the current keycloak, but not flexible and comprehensive.
 
-To realize these three points in a unified way, here "Client Policies" concept is introduced.
-
 3 will be realized by [Client Conformance Profiles](https://issues.redhat.com/browse/KEYCLOAK-11612) based on this "Client Policies". It discusses which kind of security profiles like FAPI are supported and how to implement them.
+
+4 has not yet been realized by the current keycloak.
+
+To realize these four points in a unified way, here "Client Policies" concept is introduced.
 
 ## Use-cases
 
@@ -46,9 +49,13 @@ The pre-set client policies for following some security profiles (FAPI, SPA, Nat
 It is treated in [Client Conformance Profiles](https://issues.redhat.com/browse/KEYCLOAK-11612) based on this Client Policies. Which kind of security profiles are supported and how to realize them are discussed on 
 [Client Conformance Profiles](https://issues.redhat.com/browse/KEYCLOAK-11612).
 
+## Scope
+
+This client policy concept is independent of any specific protocol. However, this document only deals with the implementation for OIDC.
+
 ## Implementation details
 
-Client Policies consists of the three building blocks, Policy, Condition, and Executor.
+Client Policies consists of the four building blocks, Policy, Condition, and Executor.
 
 ### Condition : which client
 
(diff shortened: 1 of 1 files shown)


design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+
+* The way of creating/updating a client (OIDC Dynamic Client Registration - Authenticated, Anonymous, Admin REST API - Admin Console, etc.)
+
+E.g. When creating a client, a policy is adopted when this client is created by OIDC Dynamic Client Registration with Initial Access Token.
+
+* Client Access Type (confidential, public, bearer-only)
+
+E.g. When a client sends an authorization request, a policy is adopted if this client is confidential.
+
+* Client ID
```
stianst, line 77: Not sure I see how "Client ID" is that useful as a condition. It's to specific and would make it very hard to manage policies. Client ID is also not something that is controllable - to prevent a policy being triggered a developer could just use a different client-id.

design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+
+#### How to configure
+
+A condition can be configuable the same as other configurable provider. What can be configured depends on each condition's nature.
+
+#### Which kind of conditions are provided
+
+The following conditions are provided:
+
+* The way of creating/updating a client (OIDC Dynamic Client Registration - Authenticated, Anonymous, Admin REST API - Admin Console, etc.)
```
stianst, line 69: Should probably also have some conditions on the author of the client. User/group/role of creator of the client, or different policies based on different initial access tokens, etc.

design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+
+* Client Access Type (confidential, public, bearer-only)
+
+E.g. When a client sends an authorization request, a policy is adopted if this client is confidential.
+
+* Client ID
+
+E.g. When a client sends a token request, a policy is adopted if this client's ID (`client_id` in OAuth2) is "TPP-ClientApp".
+
+* Client Redirect URI
```
stianst, line 81: Wonder if "domain names" would be better than redirect-uri. You'd specify this policies applies to "myapp.comany.com" or "company.com". Then it would scan all URLs defined by the client if they apply or not.

design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+
+E.g. When updating a client on Admin Console, a policy is adopted if this client's updating redirect URI is exactly matched with "https://www.example.com/callback/".
+
+The followings are adpoted only when creating or updating clients:
+
+* Client Role
+* Client Scope
+* Group
+* Host
```
stianst, line 90: Host of what - the author, or host in the client description?

design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+* User
+* User Role
+
+These conditions are evaluated against some properties or contexts of a client application(e.g. `security-admin-console`) and its user who is trying to create or update clients.
+
+E.g. When creating a client, a policy is adopted if it is on Admin Console and the user of this Admin Console belongs to the group "IT-Dept".
+
+### Executor : what action
+
+An executor specifies what action is executed on a client to which a policy is adopted. The executor executes one specified action. Such the action is checking whether the value of the parameter "redirect_uri" in the authorization request matches exactly with one of the pre-registered redirect URIs on Authorization Endpoint and rejecting this request if not.
```
stianst, line 101: Would be good to mention what "events" an executor can act on, and that an executor can take actions on one or more events. We should also allow executors that make changes to the config.

I can think of the following events:

1. Create client - augment phase  - an executor can modify the config in this phase
2. Create client - validation phase
3. Update client - augment phase - an executor can modify the config in this phase
4. Update client - validation phase
5. Request (multiple events here I guess?)

In 5 not sure if that's all, or if there are more nuanced events here.

We probably also need executors to be protocol specific.

design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+### Configuration
+
+Policies, conditions, executors can be configured on Admin Console. To do so, add "Client Policies" tab on Realm->Realm Settings, which means an administrator can have client policies per realm.
+
+An administrator can create new policy and add conditions and executors.
+An administrator can update their created policies by add/modify/delete conditions and executors while the administrator can not update pre-set policies.
+
+### Backward Compatibility
+
+Client Policies can take over Client Registration. However, Client Registration Policies also still co-exist. In the future, Client Registration Policies will be removed.
```
stianst, line 186: Ideal would be to somehow automatically migrate client registration policies into client policies so we can get rid of them

design/client-policies.md
tnorimat: I see, I'll remove this "Client ID" condition.

design/client-policies.md
tnorimat: I've missed this point. I'll add this automatic migration to Backward Compatibility section.

design/client-policies.md
```diff
@@ -0,0 +1,191 @@
...
+* **Status**: Draft #1
+* **JIRA**: [KEYCLOAK-13933](https://issues.jboss.org/browse/KEYCLOAK-13933)
+
+## Motivation
+
+To make it easy to secure client applications, it is beneficial to realize the following three points in a unified way.
+
+1. Setting policies on what configuration a client can have
+2. Validation of client configurations
+3. Conformance to a profile such as FAPI
```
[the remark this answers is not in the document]
tnorimat: I see,  I'll add the point you mentioned.
"4. Limiting what type of configuration a client can have based on source that is creating the client"

design/client-policies.md
tnorimat: IMO, broadly speaking, keycloak user can register/update a client by 2 way.

1. Admin REST API (used by Admin Console, etc.)
2. Dynamic Client Registration (Anonymous/Authenticated)

As for the former, the creator of the client can be specified by 

1. The user who has the authority to creating a client
2. The client (e.g. security-admin-console)

We should prepare conditions based on properties of them (e.g. user, group, role).

As for the latter, we should prepare conditions checking whether Anonymous or Authenticated registration/update.

As for Authenticated registration(with initial access token) and update(with registration access token), we should prepare conditions based on the contents of their tokens (e.g. some claim included in the initial access token). However, we need to determine which claims are used for conditions.

WDYT?

design/client-policies.md
tnorimat: I see, I'll add "Client Domain Name" condition which checks whether the registered domain name matches domain parts of all URLs of the client's properties (e.g. redirect_uris, jwks_uri).

design/client-policies.md
tnorimat: The author. When dynamic client registration, we can restrict the author of this registration by its IP address and Host.

design/client-policies.md
tnorimat: I think "events" you've mentioned are nearly equal to what I've wrote in "Endpoint and Operation" section.
Incorporating your points, I'll add descriptions about "events" in "Executor" section.

Also, I think "modify the config" you've mentioned means that the executor can modify the config of the client when creating/updating client (e.g. the author of creating the client specify `token_endpoint_auth_method=client_secret_basic` but the executor change it to `privaty_key_jwt`). Is that right?

design/client-policies.md
stianst: Dynamic client registration can also be invoked with a regular token, so probably look at it from who, not what endpoint. We have:

* Anonymous
* User - role, group, etc.
* Initial access token - Not sure what's best here, but would be good to allow creating different tokens with different access. Wonder if the ideal would just be to allow assigning for a given user/service-account or something like that? Then we're back to anonymous and user to check what policies to apply.

design/client-policies.md
stianst: Yes, not sure that's the best example, was more thinking along the lines of enabling/requiring consent. Enabling/requiring PKCE. However, there shouldn't be any limit to what it can do. Also, that means there's two phases in creation/updating. As otherwise "validation" may not be the finished config, but some interim.

tnorimat: Thanks. As you mentioned, I'll also add multi-phase action of executor in creation/updating client to "Executor" section.

design/client-policies.md
tnorimat: On dynamic client registration, I think "User" stands for Service Account because the client application itself request client registration with some kind of token.
On registration by Admin REST API, I think "User" stands for the end user like keycloak's administrator. Is that right?

As for restriction based on contents of the token used on dynamic client registration, I think it might be sufficient to use the existing properties (role, scope, etc.) .

stianst: For dynamic client registration it can also be a regular user. You may for example have some developer portal where a user can login to generate an access token that allows them to register clients.

For admin REST API, yes it would have to be an admin user.

stianst: The ideal would be if we have a single way for admin, initial access, and user, to somehow say apply this policy to this group of users, or this policy to this group of users. That does suggest that potentially an initial access token should be issued to a specific user.

tnorimat: I see, on dynamic client registration, "User" stands for the end user who was authenticated to get an access token for generating a new client, not Service Account of the existing client that actually accesses the registration endpoint with the access token.

tnorimat:
I've updated the document by incorporating comments as 1st round.

tnorimat:
Also added Token Revocation to incorporate the latest update (https://github.com/keycloak/keycloak/pull/6704).

tnorimat:
In Scope section, I've added SAML treatment recently discussed in https://groups.google.com/forum/#!topic/keycloak-dev/ZwAYoIsTEkg .

stianst:
Merging as I'm okay with it now, and no-one else has provided any input (been open for 2 weeks now)

PR #104, opened 2020-05-04
Add more details to user profile design

Martin-Idel-SI:
Opened for discussion. To be merged when deemed ready by community

### DIFF
--- a/design/user-profile.md
+++ b/design/user-profile.md
@@ -86,15 +86,37 @@ A very rough idea on how the user profile JSON could look like:
     {
         "attributes": [{
             "name": "department",
-            "view": ["admin", "user"],
-            "edit": ["admin"],
-            "requirement":"always",
+            "permissions": {
+                    "view": ["admin", "user"], 
+                    "edit": ["admin"],
+            }
+            "requirement": {
+                "always",
+            }
             "validation": {
                 "name", 
-                { "length" : { "min" : 10, "max": 20 } }
+                "context" : ["UserProfileUpdate"],
+                { "length" : { "min" : 10, "max": 20 } },
+            },
+            "converter": {
+                "datetime-converter": {
+                    "datetime-format" : "tt/mm/yyyy"
+                },
+                "timezone-converter": {
+                    "z": "UTC-0"
+                };
+            },
+            "annotations":  {
+                "key": "value",  
+                "framecolor": "red",
+                "gui-order": "1",
+                "type": "dropdown",
+                "defaults" : [1,2,3,5]
             }
-        }
+        }]
     }
+
+### Required Fields
 
 Requirement should support following values:
 
@@ -108,13 +130,15 @@ The `user` requirement allows an admin to create a user without specifying the v
 
 The `scope` requirement allows an attribute to only be required when a specific client scope is being requested.
 
+### Validation
+
 Validation should support the following values:
 
 * `validator-id`
 * `{ "validator-id" : "<config>" }`
 * `{ "validator-id" : { ... } }`
 
-This will allow using built-in validators as well as adding custom validators. A validator can take an option config either as a single field or a JSON object.
+This will allow using built-in validators as well as adding custom validators. A validator can take an option config either as a single field, or a JSON object.
 
 Built-in validators should cover:
 
@@ -128,6 +152,25 @@ Built-in validators should cover:
 
 There will be a user attribute SPI allowing custom validators to be created.
 
+Validators can also be attached to a special context meaning the validation will only be performed in that specific context.
+This allows distinguishing admin and user validation. For example, an admin may create a user without first and last name, but the user must provider these upon login.
+
+Possible contexts include:
+
+* UpdateProfile
+* UserRegistration
+* UserResource
+* Account
+* Registration
+
+### Converter (optional)
+
+Converters can be used to preprocess an attribute value before storing it.
+A converter may also affect the "read" process.
+
+### Annotations
+
+Annotations are e.g. unstructured labels which are populated to the freemarker context and can be considered for theme rendering.
 
 ## Built-in Attributes
 
@@ -206,9 +249,16 @@ All mappers that use user attributes should provide a drop-down with list of att
 
 ## Milestones
 
-### M1 SPI
+### M1 SPI, Legacy Provider and UserModel preparation/refactoring
 
-Define the SPI/API for user profile provider and user attribute validators.
+Define the SPI/API for the user profile, its attributes with validators (and optionally permissions).
+Implement simple legacy provider.
+
+Migrate to new user model where all fields like username, email, etc. are attributes
+Don't change database model: Use entity - model adapter classes to enforce legacy behaviour: all legacy fields are "mandatory"
+Use legacy provider to enforce old behaviour
+
+(Potentially, those can be split into several contributions)
 
 ### M2 Validation
 
@@ -218,7 +268,9 @@ This will require the legacy provider, and until the default provider is added u
 
 ### M3 Default Provider with additional attributes
 
-Add default provider with basic text-area to edit the profile in the admin console. Registration form, account console and admin console will be expanded depending on what is decided in the dynamic forms section.
+Add default provider with basic text-area to edit the profile in the admin console. 
+Registration form, account console and admin console will be expanded depending on what is decided in the dynamic forms section.
+Introduce client-scope based control of requirement configuration.
 
 ### M4 Mappers
 


design/user-profile.md
```diff
@@ -86,27 +86,63 @@ A very rough idea on how the user profile JSON could look like:
...
 * `optional` (default if no value given)
 * `always`
 * `user`
-* `{ "scope" : "scope-1" }`
-* `{ "scope" : [ "scope-1", "scope-2" ] }`
 
 The `user` requirement allows an admin to create a user without specifying the value, while the user will then be required to enter the value on first login.
 
-The `scope` requirement allows an attribute to only be required when a specific client scope is being requested.
+### Profile Attribute Requirement Configuration SPI
```
stianst, line 126: Can you elaborate on other use-cases than client scope here? I would like to understand the need for other use-cases and an SPI before we bring in that additional level of complexity.

Based on the example I don't understand how this would be applied to the user profile. Where does the snippet go in the profile JSON? Before we had a simple requirement field attached to each attribute.

design/user-profile.md
```diff
@@ -128,6 +164,14 @@ Built-in validators should cover:
 
 There will be a user attribute SPI allowing custom validators to be created.
 
+### Processors (optional)
```
stianst, line 167: Would "Converter" be a better name? Processor makes me think that it does more than just to conversions.

design/user-profile.md
```diff
@@ -206,9 +250,16 @@ All mappers that use user attributes should provide a drop-down with list of att
...
 ## Milestones
 
-### M1 SPI
+### M1 SPI, Legacy Provider and UserModel preparation/refactoring
+
+Define the SPI/API for the user profile, its attributes with validators (and optionally permissions).
+Implement simple legacy provider.
+
+Migrate to new user model where all fields like username, email, etc. are attributes
+Don't change database model: Use entity - model adapter classes to enforce legacy behaviour: all legacy fields are "mandatory"
```
stianst, line 259: Do you mean removing getFirstName/LastName/Email fields on UserModel?

stianst:
We should go with the dynamic option for built-in forms. Then allow adding custom crafted forms when needed. Otherwise, it would be required to modify many templates just to add a single field, which is not good.

design/user-profile.md
bs-matil: yes, we would remove it completely in the whole middle tier. It will only stay in the representation/context and the entity in the first revision we plan right now. The idea is to partially fade away from the notion of explicit fields of a user. Instead, there are some well-known attribute keys e.g. username. 

Our thought here is that in the future there will be no "column" username as there are no columns.. so specific handling in the model for this field seems not necessary. 

This would enable consistent handling of the profile and its values in the whole application as there are no static fields anymore. And we can define (hardcoded) validators for all fields also the username and email. 

Do you have concerns that the username should stay a fixed field? 
We could easily agree that e.g. username and email stay fixed as they have special meanings regards identity of a user. 
The question is would you like to be able to have validation functions "non hardcoded" on those fields. if not we can leave them static. But it would be hard to attach validators to nondynamic fields. 
But on the other hand, we would like to have mostly dynamic handling of user information.

design/user-profile.md
bs-matil: we will use Converter from now on

design/user-profile.md
bs-matil: Regards the profile JSON. The idea is to store the configuration with the element which defines it. E.g. configs for clients are defined on the client level and are also stored on the client level. 

The idea is to support any kind of profile override in a given context. Example: a custom required action would like to add mandatory fields for the profile. Then it would implement this interface and configures the fields in its config. When a user tries to use the action he will be prompted with the merged configuration. 

In general, the idea is to establish a general configuration manipulation hook for all contexts which can reasonably try to enrich the profile based on their state. For example, imagine different mandatory fields are given if you are logging in via different idps. If you log in from BusinessContextA you want other mandatory fields than in BusinessContextB but it's not decided on client level but on the login context. The idea is to support any kind of these use cases with an extensional API.

stianst: Afraid you've completely lost me here. I don't understand how this is supposed to be configured/used.

From my perspective a client needs a specific client scope, and a client scope requires the user profile to contain specific attributes.

I don't see a use-case for anything beyond that. Can you explain the use-cases you are trying to cover, and also some more details on how this is configured/used?

design/user-profile.md
```diff
@@ -86,27 +86,63 @@ A very rough idea on how the user profile JSON could look like:
...
             "name": "department",
-            "view": ["admin", "user"],
-            "edit": ["admin"],
+            "permissions": {
+                    "view": ["admin", "user"], 
+                    "edit": ["admin"],
+            }
             "requirement":"always",
             "validation": {
                 "name",
```
bs-matil, line 95: @stianst we experience some issues regarding the topic of validation.
While browsing the codebase it turns out there are several completely incompatible ideas of validation right now:
UpdateProfile --> The profile is not validated as in the registration but some assessments are done based on realm settings
UserRegistration --> validation logic server side
UserResource/AdminUI--> validation logic client side

This leads to the problem that if one merges all those logic into one place it has one of two possible consequences. 
1) Everywhere the same validation logic -> can break admin UI themes 
2) Add contexts to each validator e.g. UserProfileUpdate vs UserResourceUpdate vs UserRegistration vs ALL etc (not all known right now). This way validators can be specified for each context, which is especially useful for the legacy provider. 
The default context would be ALL and has not to be configured explicitly. 
This also implies that an attribute can have several instances of the same validator for different contexts. I don't expect to this kind of configuration in the wild but it's needed to fulfill the legacy compatibility. 

Validator config could look like this. 

```
 "validatiors": [{
                "name": { context:"UserProfileUpdate ", "length" : { "min" : 10, "max": 20 } }
                 }],
```

design/user-profile.md
stianst: I agree that most things should be attributes, not fields. Even the JSON representation at some point needs a cleanup, but that is a separate thing to this. There's quite some things that needs to be figured out though to be able to update the UserModel though.

Keycloak uses username, email and email-verified quite a lot, and needs a fixed way to access this. Further, username and email use DB constraints to make sure they are unique. Not having them as columns complicates that. Then again, emails sometimes should be unique and sometimes not. User sometimes have a username, and sometimes only an email to login with.

design/user-profile.md
bs-matil: If you abstract the problem away from scopes at all. And remove the coupling to scopes for a second it may get more clear. 

The scope topic in general terms tries to achieve manipulation of the profile configuration based on scope values. 
Therefore we need to process "some" code in case a flow happens in a specific context (a client in this case). Instead of hardcoding the scope handling, we propose to hardcode the ways the profile can be manipulated.

For example, a simple interface for the SPI could look like this
`Set<String> getMandatoryAttributes()`

This interface is called when the profile (which always gets the session context) determines that the SPI is registred for this contexts (e.g. for an authentication session for a specific client) 

This enables us to define any kind of SPI to manipulate the mandatory fields in the profile in different contexts. For example, an Identity provider implementation could set mandatory fields based on the provider config. So different providers would enforce different fields for the same user (e.g. to reflect a specific business context, or fields which are automatically set in on case but not in the other) 

The idea is to introduce a general notion of how other components than the profile are able to manipulate the config based on their context. 

I hope this makes it more explicit. If you feel that there is no general need to enforce different profile behavior based on different contexts (provider, client, account ui, resource api) we can drop this whole approach for sure.

design/user-profile.md
Martin-Idel-SI: Yes, that's something we are also aware of (and therefore we wanted to specifically address this here). Our idea was the following:

We don't want to change the database model - at least in the beginning, so the columns would stay the same (in other words: the UserEntity wouldn't be touched). We would make sure that the UserAdapter and the RealmAdapter ensure that the user fields are correctly set and handled. Then you could figure out later what to do with the database model, whether or not to remove the columns (I guess users may have indices on the username column or the like so that needs to be very carefully considered), etc.

We do however want to also switch the fields to attributes in the UserRepresentation and therefore the JSON representation, but we think this is best done later when working on the UI to reflect the profile, because it seems to me that it is more difficult to ensure this is done correctly and we'd need more familiarity with the codebase in the UI first. 

We see this as a tradeoff: 
- When modelling fields as attributes in the UserModel (and representation), this would allow a simpler and cleaner implementation of the user profile and at the same time allow us to cleanly enforce things like unique usernames and/or emails and make that behaviour configurable in the profile later on
- Incremental changes of the model leave us with an awkward intermediate state where e.g. the UserModel class doesn't know about "First name" while the representation and entity classes still do, but then smaller changes are easier to get right (and review)
- From what we gather in your design documents, you wish to migrate to this kind of user model where many things are attributes anyway, so we thought we could maybe make a step in this direction now but keeping the old model in the database

So in order to switch the model, here is how I would proceed for e.g. the last name (email and username would be slightly more difficult probably):
- Define a class like `BasicUserAttributes` or maybe `OIDCBasicClaimAttributes` or similar where we define a constant `GIVEN_NAME="given_name"` as the last name could match the corresponding open id claim.
- Replace all occurrences of `UserModel.setLastName(name)` by `UserModel.setSingleAttribute(GIVEN_NAME, name)` (and similar for the getter). 
- In the database adapter (UserAdapter) we'd rewrite the methods setting and getting attributes to add the given name to the attributes (getting attributes) or to call `entity.setLastName` if the attribute to set was `GIVEN_NAME`. That way, the last name attribute should still be saved and read from the corresponding column, but internally, it's being added to the attributes and used as if the field didn't exist. There is of course some runtime overhead to this, but I'd hope it's small enough to not matter.
- Similarly, the `RepresentationToModel` classes (and vice versa) have to map between the field in the representation and the attribute in the model until the field gets removed from the representation later on.

If you think we are missing something or would like to see an even more concrete plan, I can open a PR implementing these changes to further discuss this issue.

design/user-profile.md
stianst: Having the context of the update makes a lot of sense. Would be good to list the contexts we expect somewhere in this doc.

To add to the legacy use-case I can definitively think of other use-cases for being aware of the context.

Example:
* Admin registers account with email only - user is then asked to enter firstname/lastname on first login
* User registers themselves with email, firstname and lastname

For the legacy provider though what I had in mind was that would just be hard-coded as a legacy provider rather than using the JSON profile. It would still need to know the context though to be able to set the correct mandatory attributes as well as validation. Another way of doing that could be to have a pre-defined default JSON profile as well as a pre-defined legacy JSON profile. Current realms during migration would get the legacy profile, while new realms would get the new default profile. We should then probably have an easy option in the admin UI to switch to default, as well as reset a custom profile to the default.

design/user-profile.md
stianst: In general this sounds like a sensible plan. Probably worth dropping an email around this on the dev mailing list separately and have this done in a PR as a pre-requisite to this work.

design/user-profile.md
stianst: One core requirement here is the ability to have progressive user profiles. The idea on how to solve that would be that users can register with an initial basic set of attributes, but then as clients need more information about the user the users would be required to provide this information. This use-case is well solved by only requiring some attributes for specific client scopes, as now clients can request a default client scope to always require additional attributes, or can have optional scopes to require it only when they need it. Beyond that I don't think there's a case to configure different attribute requirements for a client.

Then there's the case of an external user authenticating which is a lot more complex. An LDAP provider may for example not have all the attributes required by a realm. What happens then? If the LDAP provider is read-only and doesn't allow storing additional things in KC DB, what do we do? Same goes to IdPs once we enable transient users (users only exists in KC during the time they are logged-in, but are not persisted to the KC DB). I can think of a lot of clever things we can do in these cases - including the ability to validate a LDAP configuration based on the user profile to make sure that all required attributes are mapped, including having drop-downs when selecting user attributes rather than free-text. We can also have options to reject logins if they don't match the required profile, ask the user to update the profile, or just ignoring it.

I can also see the use-case of having different attributes for different groups of users. For example you may require employees to provide an employment-id, while customers don't have to. This could be based on marking attributes as required if a user has a attribute/role/group.

I don't think it's a good idea to have an IdP implementation able to alter the user attributes, that would probably quickly become unmanageable and a bit of a spaghetti code mess. From my perspective the user profile is something that is defined based on requirements on user groups as well as what clients need, as such it's rather the user profile that instructs the IdP or user storage provider, not the other way around.

design/user-profile.md
Martin-Idel-SI: The plan would be to do this in a separate PR, it sounds like a good idea to announce this in a separate mailing list thread - I'll do that soon.

design/user-profile.md
bs-matil: As we also see your concerns we will go with the client only approach as discussed in the initial draft. We will roll back the changes in the PR soon.

design/user-profile.md
bs-matil: We agree to this approach, We will favor a pre-defined legacy profile if possible. Right now we think there might be edge cases which can't be overcome with the config approach (e.g. lagacy persistence logic) but we'll try to do so. 

As result it might turn out that we can only offer a "as close as it gets" config for the lagcy profile which can be used by people which want to keep the old behavoir. But this will may still have some changes in the details we have to communicate in the release updates.

Martin-Idel-SI:
@stianst I updated the PR according to the discussion here. Hopefully I caught everything.

stianst:
LGTM

PR #153, opened 2020-06-09
KEYCLOAK-10213 Store.X architecture

hmlnarik:


design/keycloak.x/storage-architecture.md
```diff
@@ -0,0 +1,129 @@
...
+referential integrity.
+
+### First-class objects
+
+First-class objects are objects that have a common store settings, and
+their separation is logical  from the perspective of objects handled in
+Keycloak. The first-class objects are depicted below in the color boxes.
+A **realm** contains all data related to the realm that almost never
+changes (realm name and other attributes, identity providers, user
+federation, flows setup of keys etc.). A **client** contains similarly
```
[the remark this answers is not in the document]
mposolda: Thanks Hynek for the clarification!

design/keycloak.x/storage-architecture.md
```diff
@@ -0,0 +1,129 @@
...
+
+![First-class objects](img/storage-first-class-objects.png)
+
+### Store configuration
+
+For each object type, a separate store can be configured. The store
+allows configuring persistence and caching independently. For the current
+store, JPA would be used for persistence and Infinispan for caching in
+case of realms, roles etc., and Infinispan would be used for persistence
+and passthrough (i.e. nothing) for caching in case of sessions.
```
slaskawi, line 69: What about cross-site replication? I don't see it mentioned anywhere, yet it needs to somehow fit into the picture. 

Since we have the caching and the storage layer, where would you plug the x-site in?

design/keycloak.x/storage-architecture.md
```diff
@@ -0,0 +1,129 @@
...
+   thus depend on the state of the objects (realms, sessions etc.)
+   throughout the period of handling the request
+
+2. Stateless representation (similar to aggregated `*Entity`) - the
+   representation contains IDs rather than ability to provide logical
+   representation of the object. The crucial feature of each field type is
+   that its representation is serializable, and is independent of the
+   datastore format.
+
+3. Storage representation - datastore-dependent representation, e.g.
```
slaskawi, line 98: What will the be "future" workflow if I wanted to add a new field to the Realm? What objects will I need to modify?

From my understanding, I would need to add it to the logical representation and all storage representations as well (Infinispan, JPA, YAML, and others). How do we plan to keep all this places consistent?

design/keycloak.x/storage-architecture.md
hmlnarik: Both will be affected - caching at least by invalidation events, storage by defining sharing / sharding policy.

design/keycloak.x/storage-architecture.md
hmlnarik: This is a very good question for an actual store. You should need to modify the model and a common ancestor, that should be it.
For storage, it will depend on the level of the particular storage implementation, but every storage will have support for storing unknown fields in some form, e.g. in JSON, and also all the fields will need to be serializable to a String for the same reason. If desirable, the storage can adjust physical layout to provide the best fit for a particular field, e.g. by storing its value in a column in a table in case of JPA.

PR #154, opened 2020-06-10
Configuration design for Keycloak.X

stianst:


design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+The aim is to start small with what config options we provide, and extend when there is a need. Great care should be 
+taken in reducing the amount of configuration needed, where opinionated/best-practice behaviour should be favoured, with
+high-level and simple configuration options introduced only when needed.
+
+Below is a list of config by category we should introduce initially:
+
+### HTTP
+
+Keycloak.X will not come with the capability to generate a self-signed certificate. Instead, Keycloak by default will
+complain if a certificate has not been configured, but will allow explicitly disabling SSL for development purposes.
```
sschu, line 101: This might be understood as suggesting plain HTTP is only OK for development. However, setups with reverse proxy and/or sidecar might also be valid for production.

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+
+Setting `proxy.mode` will under the covers configure `http.mode` as well as `quarkus.http.proxy-address-forwarding`.
+
+## Database
+
+The aim is that a user should not need to understand Hibernate, JDBC drivers, etc. to be able to configure the database.
+As such we should introduce a high-level option `db.vendor` instead of requiring users to set JDBC driver and Hibernate
+dialects.
+
+We will support either `db.url` which will support a JDBC URL, or using separate properties to build-up the URL behind the
```
sschu, line 133: Does "either" mean it is not decided yet or "both"? If only construction is possible, it should also be supported to add custom parameters to the URL since there are so many different parameters for the different databases.

sschu: Reading this a second time, it probably means both. Then I would really like to have a db.custom_params configuration property.

design/keycloak.x/configuration.md
stianst: See the section on proxy mode

design/keycloak.x/configuration.md
stianst: The thinking was to support both. Big question really is what is most convenient, also it is slightly tricky to let you "build" the JDBC URL in an intuitive way since there isn't much of a standard there AFAIK, where all vendors have different syntax. So perhaps the "builder" approach is for novice users, while the direct URL approach is for power users.

thomasdarimont:
For the record this is what I currently do in the embedded-spring-boot-keycloak server:
https://github.com/thomasdarimont/embedded-spring-boot-keycloak-server/blob/master/embedded-keycloak-server-custom/src/main/resources/application.yml#L33

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+Keycloak.X can be configured in a property file, using command-line arguments or environment variables. All configuration
+properties are available in all different approaches.
+
+Properties (conf/keycloak.properties):
+
+    hostname.frontendUrl=https://mykeycloak
+    
+Environment variables:
+
+    export KC_HOSTNAME_FRONTEND-URL=https://mykeycloak
```
slaskawi, line 16: We should probably unify underscores (`_`) and hyphens (`-`). Spring uses a pretty nice convention ([link](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-relaxed-binding-from-environment-variables))

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+  
+## SPIs and Provider Configuration
+
+Properties for providers follow the following format:
+
+    spi.<spi-name>.<provider-name>.<config-name>
+
+For example:
+
+    spi.hostname.default.frontendUrl=https://mykeycloak
```
thomasdarimont, line 54: If you have multiple providers for the same spi, how do you specify which one should be used by default?

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+    
+## Alias
+
+It should not be required to use SPI/provider specific property names for commonly configured properties for default
+providers. Here we will support alias for properties.
```
thomasdarimont, line 59: Is that to remain compatible with the existing Keycloak SPI provider configuration?

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+* db.username
+* db.password
+* db.pool.initialSize
+* db.pool.minSize - default to 0
+* db.pool.maxSize - default to 100
+
+## Caches (Infinispan and JGroups)
+
+As we aim to not use an embedded Infinispan server in the future, configuration of caches will not for now follow the
+standard way of configuring Keycloak. Instead there will be a separate XML file for configuring this (`conf/cluster.xml`).
```
arndt-s, line 151: I'd prefer a uniform format for the configuration. Currently caches would be configured via xml, the others via .properties files.

design/keycloak.x/configuration.md
stianst: I think the options are to just convert camelCase to CAMELCASE, or do CAMEL-CASE. I probably perfer CAMELCASE to be honest. Doing CAMEL_CASE is not a good approach as there's then no differentiation between "my.name" and "myName" when converted.

design/keycloak.x/configuration.md
stianst: The thinking was `spi.hostname.provider=mycustom`. It does mean you can't have a provider named `provider` as you'd have a name clash, but I think that's unlikely and can't think of a cleaner way to do it.

design/keycloak.x/configuration.md
stianst: Not really. We should have alias/high-level properties for common configuration things, then have the spi/provider specific properties for less common things, and custom providers (custom providers shouldn't use high-level/alias as that would then easily end up with conflicting with KC specific ones).

design/keycloak.x/configuration.md
stianst: I appreciate that, but as stated the plan is to not use an embedded Infinispan server in the future, which would make cache configuration much simpler (would be closer to how you configure a DB with just specifying the url, not stacks, etc.).

We may be able to come up with some slightly nicer interim solution though.

thomasdarimont: Did you see this? Referencing an jgroups config from infinispan config:
https://github.com/keycloak/keycloak-community/pull/154#issuecomment-645234340
Perhaps this could be the interim solution

stianst: I think @pedroigor proposed something similar. I'd like to split this up though. First pass would be to just make caching configerable, then we can look into usability improvements around it later.

design/keycloak.x/configuration.md
pedroigor: What if we do:

```
# maps to the default
hostname.frontendUrl=https://mykeycloak

# maps to a specific provider
hostname.default.frontendUrl=https://mykeycloak
hostname.custom.frontendUrl=https://mykeycloak
```

The logic to map to the correct provider would be based on checking whether or not the part after the SPI name can reference a provider.

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+taken in reducing the amount of configuration needed, where opinionated/best-practice behaviour should be favoured, with
+high-level and simple configuration options introduced only when needed.
+
+Below is a list of config by category we should introduce initially:
+
+### HTTP
+
+Keycloak.X will not come with the capability to generate a self-signed certificate. Instead, Keycloak by default will
+complain if a certificate has not been configured, but will allow explicitly disabling SSL for development purposes.
+
```
pedroigor, line 102: What about:

```
https.enabled=true|false
http.port=8080
https.port
https.certificate.*
```

```
https.enabled=false
http.port=8080
```

I think this approach:

* Removes the `ssl` part making the property easier to type and read
* Explicitly shows that we care about `https` which is enabled by default.

design/keycloak.x/configuration.md
pedroigor: Thinking more about it, I think we should not use abbreviations but `database`. It looks weird to me and not so friendly if we just use `db`.

It is also aligned with other examples where the correct name (e.g.: certificate and not cert|crt) is being used.

There are also some other properties that we may want to expose here but we can always document them later, like enable metrics.

design/keycloak.x/configuration.md
pedroigor: Currently, users have complete control over the `cluster.xml` file so they can configure ISPN at their weill. Including doing something like that, @thomasdarimont.

However, I see as a good improvement to support defaults to make life easier for people running on well-known cloud providers or on-premise:

```
cluster.ec2.enabled
cluster.ec2.s3.region-name
```

```
cluster.kubernetes.enabled
cluster.kubernetes.dns.query
```

```
cluster.baremetal.tcp.hosts
cluster.baremetal.udp.multicast.address
```

```
# For Cross-DC on top of a particular stack something like
cluster.kubernetes.enabled
cluster.kubernetes.dns.query
cluster.cross-dc.enabled
cluster.cross-dc.option1
```

And so forth, where most of the configuration above would come from the defaults shipped with JGroups itself. Plus any other default that we make sense for running Keycloak on top of those platforms.

For more advanced use cases and for those looking to hack their own configuration, `cluster.xml`. 

@mposolda Can help us a lot here ...

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+## Caches (Infinispan and JGroups)
+
+As we aim to not use an embedded Infinispan server in the future, configuration of caches will not for now follow the
+standard way of configuring Keycloak. Instead there will be a separate XML file for configuring this (`conf/cluster.xml`).
+
+## Logging
+
+TBD
+
+## Metrics
```
pedroigor, line 157: I would say we need a separate document for this one so that we can come up with a nice solution that leverages MP Metrics that make easier setup richer and meaningful Grafana dashboard from the default metrics we provide, for instance.

design/keycloak.x/configuration.md
slaskawi: I agree with aligning with whatever Quarkus does.

stianst: Changing to _

design/keycloak.x/configuration.md
stianst: Are you saying to remove `spi` part? That's there to prevent collisions, as we could end up having SPIs with a name that conflicts with other "top-level" categories.

I don't understand what you mean about "The logic to map to the correct provider would be based on checking whether or not the part after the SPI name can reference a provider"

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,159 @@
...
+* All options must be in a category (`<categoryName>.<propertyName>`)
+* Multiple levels of categories are permitted (`<categoryName>.<subCategoryName>.<propertyName>`)
+
+Rules for converting property keys to different formats follow these rules:
+
+* Environment variables: 
+  - Prefix with `KC_`
+  - Characters converted to uppercase
+  - Dot (`.`) replaced with underscore (`_`)
+  - Use hyphen (`-`) for camel-case names (`PROPERTY-NAME`)
```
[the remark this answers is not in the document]
stianst: Changing to CAMEL_CASE

design/keycloak.x/configuration.md
stianst: Yep, hence it's TBD

design/keycloak.x/configuration.md
stianst: +1 Doing this

stianst:
Update the proposal, a quick summary:

* Avoid camelCase everywhere - Quarkus doesn't use camelCase and we use an unhealthy mix. I propose should have a convention to not use camelCase
* Added config for profile, including a dev-mode
* Changed http.ssl to https
* Added provider and enabled to SPI section
* Updated database section
* Changed cluster section to TBD - discussion around clustering should be done separate to the initial config proposal since it will be a big topic on its own

design/keycloak.x/configuration.md
stianst: Closing this as I want to defer discussion around caches to a follow-up

design/keycloak.x/configuration.md
pedroigor: Not removing the `spi` part. But just avoiding `spi.hostname.provider` to tell which provider to configure.

stianst: I'm not really following what you are proposing then ;)

thomasdarimont: how about:

# Provider configuration
```
hostname.default.frontendUrl=https://mykeycloak
hostname.custom.frontendUrl=https://mykeycloak
hostname.custom.foo=bar
```

# Provider selection

## If you want the default, use:
```
spi.hostname=default
```

## If you want the custom, use: (also via -Dspi.hostname=custom)
```
spi.hostname=custom
```

pedroigor: Looks better.

My proposal was based on the assumption that you wouldn't need to set the provider id/alias as you only have a single one defined in your configuration. Like for `hostname', you would have either `default` or `custom`, right? So, based on this assumption we wouldn't need to set the provider id/alias. Not sure if this always makes sense.

pedroigor:
I've started the necessary work to change the packaging format to a mutable jar so that we can leverage the build time optimizations done by Quarkus when changing things such as the database.

This package format has a direct impact on how you configure the server where differently than today, you need an intermediary step to re-build the generated artifacts with the changes to the configuration (build time config like the database).

The same goes in regards to deploying custom providers, where today we configure then at runtime whereas using this new packaging format we should be able to re-build the application so that any custom provider is configured at build time, thus giving a better startup and footprint given that they are configured at build time.

That said, I'm wondering if we should consider in this initial proposal the constraints and improvements around the mutable packaging format and how the configuration should work.

pedroigor:
Initial changes to re-augmentation here https://github.com/keycloak/keycloak/pull/7207. I hope you have a better picture about this work impact configuration and possibly this initial proposal.

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,193 @@
...
+* Multiple levels of categories are permitted (`<category-name>.<sub-category-name>.<property-name>`)
+* Avoid camelCase
+
+Rules for converting property keys to different formats follow these rules:
+
+* Environment variables: 
+  - Prefix with `KC_`
+  - Characters converted to uppercase
+  - Dot (`.`) replaced with underscore (`_`)
+  - Use hyphen (`_`) for camel-case names (`PROPERTY_NAME`)
```
mingjliu9, line 43: Is `hyphen` a typo here?

design/keycloak.x/configuration.md
stianst: I think it needs to be explicit as there could be no configuration for the provider you want to use as default.

stianst: `spi.hostname=default` is shorter and kinda nicer, but it would be problematic if for some weird reason we wanted to support a structured format like JSON in the future (don't think we would though). Still not 100% sure about it though. Actually thinking it might be better to make it even more explicit like `spi.hostname.defaultProvider=myprovider`. That's kinda fuggly though.

design/keycloak.x/configuration.md
stianst: Sure is. Updated it.

stianst:
> Initial changes to re-augmentation here [keycloak/keycloak#7207](https://github.com/keycloak/keycloak/pull/7207). I hope you have a better picture about how this work impact configuration and possibly this initial proposal.

Thanks, we need to consider the impact on the design before deciding on a final approach.

I'm thinking we would have commands in "kc.[sh|bat]" with something like:

```
kc.sh update-config --database.vendor=mariadb
kc.sh start --database.url=...
```

Default command if none is specified could be start. So the following would be good as well:

```
kc.sh update-config --database.vendor=mariadb
kc.sh --database.url=...
```

We will need to clearly document what config properties require a "config update", and which can be specified at the start. Perhaps we could even do something like:

```
kc.sh --help  # Displays commands, and also describes how to list config properties using update-config --help and start --help
kc.sh update-config --help #Displays options that need config update
kc.sh start --help #Displays options that can be specified at runtime without config update
```

With commands we can also add nicer way to do import/export, including the ability to do import/export without starting KC. So something like:

```
kc.sh export --directory mydir
kc.sh import mydir
```

pedroigor:
> > Initial changes to re-augmentation here [keycloak/keycloak#7207](https://github.com/keycloak/keycloak/pull/7207). I hope you have a better picture about how this work impact configuration and possibly this initial proposal.
> 
> Thanks, we need to consider the impact on the design before deciding on a final approach.
> 
> I'm thinking we would have commands in "kc.[sh|bat]" with something like:
> 
> ```
> kc.sh update-config --database.vendor=mariadb
> kc.sh start --database.url=...
> ```
> 
> Default command if none is specified could be start. So the following would be good as well:
> 
> ```
> kc.sh update-config --database.vendor=mariadb
> kc.sh --database.url=...
> ```
> 
> We will need to clearly document what config properties require a "config update", and which can be specified at the start. Perhaps we could even do something like:
> 
> ```
> kc.sh --help  # Displays commands, and also describes how to list config properties using update-config --help and start --help
> kc.sh update-config --help #Displays options that need config update
> kc.sh start --help #Displays options that can be specified at runtime without config update
> ```
> 
> With commands we can also add nicer way to do import/export, including the ability to do import/export without starting KC. So something like:
> 
> ```
> kc.sh export --directory mydir
> kc.sh import mydir
> ```

All that looks good to me. For me, it is a great start.

I would just change `update-config` to `config` (the `configure` option I'm using that PR is too long).

design/keycloak.x/configuration.md
stianst: I'm sticking with `spi.hostname.provider` as it is not that commonly set, it's more clear and avoids any issues with conflicts in structured config formats if we want to include that in the future.

design/keycloak.x/configuration.md
dteleguin: > However, I see as a good improvement to support defaults to make life easier for people running on well-known cloud providers or on-premise:
> 
> ```
> cluster.ec2.enabled
> cluster.ec2.s3.region-name
> ```
> 
> ```
> cluster.kubernetes.enabled
> cluster.kubernetes.dns.query
> ```

Probably not the best idea to provide the `cluster.*.enabled` options; thus, we will let the user "enable" multiple JGroups configs at once, which doesn't make sense. I'd rather suggest something like this:
```
cluster = ec2
# or maybe "cluster.stack = ec2", or "cluster.type = ec2"
cluster.ec2.foobar = ...
```

> 
> And so forth, where most of the configuration above would come from the defaults shipped with JGroups itself. Plus any other default that we make sense for running Keycloak on top of those platforms.

AFAIK, JGroups do not ship the defaults, but Infinispan [does](https://github.com/infinispan/infinispan/tree/master/core/src/main/resources/default-configs). It would be nice to have an ability to activate those defaults right from the Keycloak config.

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,264 @@
...
+to get the list of static and dynamic properties. Something like:
+
+```
+Usage: kc.sh COMMAND [OPTIONS]
+
+Commands:
+  run       Starts Keycloak
+  config    Updates configuration that requires a rebuild of Keycloak
+  export    Exports data in the database
+  imports   Imports data into the database
```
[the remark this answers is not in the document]
stianst: Thanks, fixing this one

stianst:
Added section on property replacement, and properties for layout/directories (keycloak.config-file, etc.)

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,264 @@
...
+* Use hyphen `-` to combine multiple word names (`property-name`)
+* Use dot (`.`) for separation
+* All options must be in a category (`<category-name>.<property-name>`)
+* Multiple levels of categories are permitted (`<category-name>.<sub-category-name>.<property-name>`)
+* Avoid camelCase
+
+Rules for converting property keys to different formats follow these rules:
+
+* Environment variables: 
+  - Prefix with `KC_`
```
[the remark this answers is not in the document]
stianst: I don't see any issue with using `KC_` and it's shorter, so would stick with that unless there's some objections.

design/keycloak.x/configuration.md
```diff
@@ -0,0 +1,264 @@
...
+* database.pool.max-size - default to 100
+
+* database.url - JDBC URL
+
+Setting the db.vendor will set a default JDBC URL with property expressions allowing setting parts of the URL with 
+individual properties.
+ 
+Below are the default JDBC URL values that will set if `database.url` has not been configured:
+
+* h2 - `jdbc:h2:${database.url.path:${keycloak.conf.dir}/db}` - use `database.url.path=mem:kc` for in-memory
```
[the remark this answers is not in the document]
stianst: Yes, updated

PR #164, opened 2020-08-26
Notes on how to document the new Account REST API

abstractj:
Even though it's clear that we should have everything under the REST API guidelines, I would like to remove the clutter and add the notes of my findings around the REST API documentation. So we can agree on what to do next, update the content of this document and eventually merge it with REST API guidelines.

@stianst please let me know what is missing for you or should change. The content is pretty straightforward.

### DIFF
--- a/design/rest-api-documentation.md
+++ b/design/rest-api-documentation.md
@@ -102,17 +102,22 @@ paths:
 
 As we can see, with minimal effort it is possible to produce a JSON document that describes the functionalities of our API and can be used as a foundation for UI tools like Swagger to visualize and interact with our APIs. This document is served as a static file.
 
-## Swagger UI
+## Generating static HTML files
 
-Swagger UI allows developers to visualize and generate documentation based on the OpenAPI specification. There are several ways of doing this:
+Once we provide the OpenAPI document, any tool should be able to generate documentation based on it.
+Considering usability and how simple is to generate/deploy documentation, the recommended choice to generate documentation file is [ReDoc](https://github.com/Redocly/redoc#some-real-life-usages).
+The same tool used by [Apicurio](https://www.apicur.io/) and other [companies](https://github.com/Redocly/redoc#some-real-life-usages).
 
-1. Inserting the URL in the [Swagger generator](https://generator.swagger.io/) which is available online. For example: http://localhost:8080/auth/openapi
-2. Spinning up a Docker image: 
+Generate static HTML files with redoc-cli is pretty straightforward:
+
 ```
-docker pull swaggerapi/swagger-ui
-docker run -p 80:8080 swaggerapi/swagger-ui
-```   
-3. [Swagger Codegen](https://github.com/swagger-api/swagger-codegen#building). The best way to generate static HTML files for documentation.
+npm i -g redoc-cli
+redoc-cli bundle http://localhost:8080/openapi
+```
+
+## Another alternatives 
+
+Swagger Generator is one of the several tools online to visualize and generate documentation based on the OpenAPI specification. Accessing [Swagger generator](https://generator.swagger.io/) in the browser, people should be able to insert `http://localhost:8080/auth/openapi` and see the live documentation.
 
 ## Implementation plan
 
@@ -130,4 +135,7 @@ The items below were included as nice to have because it's fairly easy to get th
 
 * Generate static HTML files for the REST API
 * Generate static OpenAPI document and add to the release distribution
+
+### Acknowledgements
   
+Thanks [Eric Wittmann](https://github.com/EricWittmann) for all the suggestions.
--- a/design/rest-api-documentation.md
+++ b/design/rest-api-documentation.md
@@ -1,0 +1,133 @@
+# REST API documentation
+
+* **Status**: Notes
+* **JIRA**: 
+  * [KEYCLOAK-9655](https://issues.jboss.org/browse/KEYCLOAK-9655)
+  * [KEYCLOAK-14744](https://issues.jboss.org/browse/KEYCLOAK-14744)
+
+
+## Motivation
+
+The value of an API to a large extent depends on how easy it is to consume. Documentation plays an important part in whether an API is useful. The main issue with the Keycloak documentation these days is to keep it up to date with the changes into the code, forcing developers to dig into the sources sometimes to understand how the API works.
+
+## Microprofile OpenAPI
+
+The Microprofile OpenAPI specification defines a standard mechanism to document REST endpoints using annotations or a pre-generated JSON. That was introduced since *WildFly 19* making things easier to be consumed by [Swagger](https://swagger.io/) tooling. 
+
+To document the external API that's exposed by the [Account REST endpoint](https://github.com/keycloak/keycloak/blob/9369c7cf4d2227ae8be40a6129ca21200ae16a9d/services/src/main/java/org/keycloak/services/resources/account/AccountRestService.java#L464-L517), we need to update the source code:
+
+```
+public class AccountRestService {
+
+    ...
+
+  @Path("/applications")
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @Operation(description = "List of applications")
+  @APIResponses(value = {
+  	@APIResponse(
+  			responseCode = "200",
+  			description = "Successful, List of applications"
+      ),
+  	@APIResponse(
+  			responseCode = "401",
+  			description = "Unauthorized, Failure"
+      ),
+  	@APIResponse(
+  			responseCode = "500",
+  			description = "Error in getting the list of applications"
+      )
+  })
+  public Response applications(@QueryParam("name") String name) {
+  	...
+  }
+
+}
+```
+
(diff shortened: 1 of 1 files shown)


design/rest-api-documentation.md
```diff
@@ -0,0 +1,133 @@
...
+
+The current Keycloak distribution still use legacy feature packs from WildFly. In order to provide a subsystem xml file so that it can be used, we need the latest changes from [introduced in WildFly 20.0.1.Final](https://issues.redhat.com/browse/WFLY-13766).
+
+2. Introduce Microprofile OpenAPI subsystem in the Keycloak distributions
+3. Update the new Account REST API with the proper annotations
+4. Make sure that nothing is broken with the Quarkus distribution
+
+### Future work items
+
+The items below were included as nice to have because it's fairly easy to get these documents out of date and introduce it requires more discussion.
```
stianst, line 129: I don't understand this comment. Do you mean if they are not generated, but rather some how hand crafted?

design/rest-api-documentation.md
```diff
@@ -0,0 +1,133 @@
...
+
+* Generate static HTML files for the REST API
+* Generate static OpenAPI document and add to the release distribution
```
stianst, line 132: Is there a Maven plugin available that can generate the OpenAPI document without starting the server?

design/rest-api-documentation.md
```diff
@@ -0,0 +1,133 @@
...
+
+2. Introduce Microprofile OpenAPI subsystem in the Keycloak distributions
+3. Update the new Account REST API with the proper annotations
+4. Make sure that nothing is broken with the Quarkus distribution
+
+### Future work items
+
+The items below were included as nice to have because it's fairly easy to get these documents out of date and introduce it requires more discussion.
+
+* Generate static HTML files for the REST API
```
stianst, line 131: Is there a decent Maven plugin available that can generate HTML documentation from the OpenAPI document? It does seem like this would be incredibly useful as I doubt everyone will want to download Swagger and generate the docs themselves.

design/rest-api-documentation.md
```diff
@@ -0,0 +1,133 @@
...
+A new dependency must be introduced in the Keycloak `pom.xml` file:
+
+```
+<dependency>
+    <groupId>org.eclipse.microprofile.openapi</groupId>
+    <artifactId>microprofile-openapi-api</artifactId>
+</dependency>
+```
+
+By spinning up the Keycloak server, developers should be able to send an HTTP request to http://localhost:8080/auth/openapi and get the OpenAPI document. Here is an example:
```
stianst, line 58: Should it not rather be '/auth/openapi/account'? We'll want to have separate OpenAPI documents for different APIs, and eventually we will want this for the admin endpoints (and others) as well.

What about versioning? Would it be '/auth/openapi/account' for the latest, and then you can also grab for a previous version using '/auth/openapi/account/<version>'?

EricWittmann: Note that Microprofile OpenAPI does not support multiple separate APIs.  The specification dictates that a single OpenAPI file be accessible at the `/openapi` endpoint (in the case of Wildfly it would be `/context/openapi` I assume, although I haven't looked at how Wildfly handles this).  Like all Microprofile specs, the MP-OpenAPI spec was meant to be used in single-application platforms.  So if you have multiple separate APIs in the same application context, and you want to document them as separate OpenAPI files, the Microprofile-OpenAPI spec will not let you do that.

That said, I do think using the MP-OpenAPI annotations as described above is the right thing to do.  It just might be the case that you will need to figure out how to leverage them in a non-standard way).

design/rest-api-documentation.md
```diff
@@ -0,0 +1,133 @@
...
+## Swagger UI
+
+Swagger UI allows developers to visualize and generate documentation based on the OpenAPI specification. There are several ways of doing this:
+
+1. Inserting the URL in the [Swagger generator](https://generator.swagger.io/) which is available online. For example: http://localhost:8080/auth/openapi
+2. Spinning up a Docker image: 
+```
+docker pull swaggerapi/swagger-ui
+docker run -p 80:8080 swaggerapi/swagger-ui
+```
```
EricWittmann, line 114: This is a very popular tool and a reasonable thing to use, certainly.  However in my opinion the user interface is ... not great.  Before making a decision on this, take a look at [ReDoc](https://github.com/Redocly/redoc) as an alternative.  We use that in Apicurio Studio to visualize the API design.  An example:  https://studio-ws.apicur.io/sharing/76f236d3-1312-4548-a4d1-400487188c66

Functionally they are very similar, I just think the user experience for ReDoc is **much** better.  The main difference between the two (functionally) is that Swagger UI is interactive - when configured properly it will allow users to not only view documentation but also make test API calls.  Useful when your Keycloak server is running and accessible, but not when publishing API docs to a project site.

design/rest-api-documentation.md
```diff
@@ -0,0 +1,133 @@
...
+
+Swagger UI allows developers to visualize and generate documentation based on the OpenAPI specification. There are several ways of doing this:
+
+1. Inserting the URL in the [Swagger generator](https://generator.swagger.io/) which is available online. For example: http://localhost:8080/auth/openapi
+2. Spinning up a Docker image: 
+```
+docker pull swaggerapi/swagger-ui
+docker run -p 80:8080 swaggerapi/swagger-ui
+```   
+3. [Swagger Codegen](https://github.com/swagger-api/swagger-codegen#building). The best way to generate static HTML files for documentation.
```
EricWittmann, line 115: I would actually recommend this instead:  https://github.com/OpenAPITools/openapi-generator

Although again, I think ReDoc is a better experience for users consuming the documentation, but that's a matter of preference (ReDoc has its own CLI for generating static/publishable HTML).

design/rest-api-documentation.md
EricWittmann: Yes:  https://openapi-generator.tech/docs/plugins/

For ReDoc I don't think there is a maven plugin, but it's an NPM based tool - so using one of the node maven plugins (e.g. `frontend-maven-plugin`) is an easy option.

design/rest-api-documentation.md
EricWittmann: Not to my knowledge.  But I think there is an opportunity here to collaborate with the SmallRye OpenAPI project (Red Hat's implementation of the Microprofile-OpenAPI spec) to create one.  I believe it would be quite easy to integrate SmallRye OpenAPI into a maven plugin to generate the OpenAPI at build time rather than runtime.  Some of the Microprofile Specification features assume things are happening at runtime (e.g. filtering, model reader, etc) but the most important part (generating an OpenAPI document from annotated JAX-RS code) can be done at build time.  All we need in the SmallRye OpenAPI impl is a ShrinkWrap `Archive` with the compiled, annotated JAX-RS classes (and java beans).  So I think a maven plugin would be fairly trivial to write.

design/rest-api-documentation.md
abstractj: @stianst I can do some updates in this paragraph. But in summary, I'd like to split the work into 2 steps if we agree:

1. Dynamic documentation. 

That's the most important thing at the moment from my point of view. And that means that changes into the code should immediately be reflected in the OpenAPI document. 

2. Static documentation

Important, but we will only generate it at each release. That means that changes between releases won't be reflected immediately and in theory, could get out of date. That's what I meant, but of course, it's not the end of the world.

The reason why I see 1 as something more important, it's because it requires changes on Keycloak that depends on WildFly and requires more testing. And also I didn't want to introduce too many changes in one go.

design/rest-api-documentation.md
abstractj: Not that I know. Maybe what Eric suggested is something to consider in the future or discuss with Quarkus team.

design/rest-api-documentation.md
abstractj: @stianst it is not necessary to download Swagger. Once we have the OpenAPI document people can use any tool to view the documentation. Of course, from the usability standpoint is not ideal, but I think that generate the OpenAPI document should be our priority. 

@EricWittmann thank you, I will give that a try.

design/rest-api-documentation.md
abstractj: Agree with Eric here. What I had in mind thinking about `admin` vs `account` endpoints make use of tags. For example:

```
@Tag(name="Account REST API", description="Account REST API operations."),
@Tag(name="Admin REST API", description="Admin REST API operations."),
```

I can do some updates into the document to cover these scenarios.

design/rest-api-documentation.md
abstractj: Thanks Eric, I will look into that and update the document.

design/rest-api-documentation.md
EricWittmann: +1 to the idea that the generation of the OpenAPI doc being the priority.  Everything else flows well once that document exists.  The trick is how/when to generate that document.

abstractj:
@stianst updated with Eric's suggestion. Indeed ReDoc should be the way to go.

PR #169, opened 2020-09-08
Review configuration

pedroigor:


pedroigor:
@stianst @mposolda There is one thing in regards to feature enablement that I would like to share. Right now we are including all providers that are not enabled by default (such as tech preview ones) into the default distribution.

IMO, we should have the main distribution with only the main providers and features enabled so that users can have a smaller distribution (and startup + footprint) as well as be able to enable features (all of them within a specific group such as preview or a specific one) when running the `config` command.

Wdyt ?

design/keycloak.x/configuration.md
```diff
@@ -59,23 +67,35 @@ For example:
     
 To override the default provider for a SPI use `spi.<spi-name>.provider=<provider-name>`.
 
-To enable/disable a provider use `spi.<spi-name>.<provider-name>.enabled=true|false`.
+The `spi` prefix can also be omitted so that you can just configure providers using:
```
mposolda, line 70: Few points regarding the configuration format of `spi.<spi-name>.<provider-name>.<config-name>` and others like `spi.<spi-name>.<provider-name>.enabled` .

Our current names of SPI, providers and configuration options currently contains all combinations of camel cases, hyphens, dots, underscores and uppercase-only formats.
Example of SPI names we have: `actionTokenHandler`, `client-registration-policy`
Example of Provider names we have: `keycloak-oidc-jboss-subsystem`, `forceExpiredPasswordChange`, `requesting_party_token`, `A128CBC-HS256`, `RSA1_5`
Example of provider config properties we have: `group.object.classes` and every other format from above

We currently don't have any convention for SPI names, provider names and configuration property names.  We may introduce it, but for now, I assume that we will need to support what we have today, which means some flexibility might be needed in evaluating property names etc. 

For example if there is environment variables like `KC_SPI_FOO_BAR_BAZ_CAS_DOO`, we may need to evaluate:
- If SPI named `foo` or `FOO` exists
- If not, we may try SPI names like `foo-bar`, `fooBar`, `foo_bar`, `foo.bar`, `FOO_BAR`, `FOO-BAR`, `FOO.BAR`
- If not, try combinations with 3 words etc

pedroigor: Good point. When I first read your comment I froze :) Good news is that you should be able to pass arguments like below and still have them fetched by the respective provider with the current implementation:

```
--client-registration-openid-connect-static-jwk-url=http://c.jwk.url
--action-token-handler-verify-email-some-property=test
```

In practice, neither of these provides support additional properties from what I could see, but if they do, they can still get the configuration without changing code.

That is due to the fact that it is easier to respect the format when providers are reading the config than beforehand when parsing the properties without any context.

mposolda: Yes, good point. So if we can count with the fact that providers are reading the config and we won't have use-cases like "parsing the properties without any context", then we should be probably fine.

PR #173, opened 2020-09-15
Minor updates to Keycloak.x config

pedroigor:


design/keycloak.x/configuration.md
```diff
@@ -373,11 +381,33 @@ By default only supported features are enabled. In addition there are some suppo
...
+
+      kc.[sh|bat] show-config
+
+The example above is going to print only the configuration set into the server image without start the server.
+
+The `show-config` command aims to help troubleshooting and investigations in case the server is not behaving or not configured as expected.
+
+The `start` method also supports using this command as a sub-command:
+
+      kc.[sh|bat] start show-config
```
stianst, line 406: Do we really need show-config as a sub-command?

bs-matil: We have a similar command in our tooling (ci/cd) and it helped us a lot. I would vote for it. It can also be used to automate tests agains the config.

stianst: It's confusing to have two commands being used, and we don't have the concept of a "sub-command" anywhere else. I'd say if we want the ability to show config and start at the same time it should be done with supporting multiple commands, like "kc.sh config start", which should then be possible for everything. Or, we can have an arg to start command "kc.sh start --show-config".

pedroigor: Agree. Will change that on both sides (dist and here).

@bs-matil good to know.

pedroigor:
@stianst It should be updated. Please, let me know your thoughts on how the document looks like now.

PR #177, opened 2020-09-24
KEYCLOAK-8288 Observability #2

slaskawi:
https://issues.redhat.com/browse/KEYCLOAK-8288

This Pull Request contains Observability design proposal. If you'd like to view it with inlined images, please use this link: https://github.com/slaskawi/keycloak-community/blob/KEYCLOAK-8288-Observability/design/observerability-2.md


bs-matil:
I really like this new proposal.

One thing we do may be interesting here too.  

We separate the view on the API levels.
so -> how healthy is the account API vs how health the admin API vs the authentication/authorization APIs.

Why? We judge different API's less critical to the customer. E.g. if the admin API fails it mostly does not affect the end-user while an issue in the login flows and authorization endpoints will affect the whole userbase. 

Does this make sense? A simple view on this would be additonal series displayed in the error graphs by api

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
+# Observability #2
+
+This document deprecates [the previous version of Observability](observability.md)
```
stianst, line 3: Replace observerability.md instead of creating a new file (you may want to keep some of the resources from the original file in an appendix/resources in the updated file

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+What's in the scope of this design:
+- Designing a new Healcheck API. This has been covered in a separate design document. See [KEYCLOAK-12398](https://issues.redhat.com/browse/KEYCLOAK-12398)
+- Designing and implementing a dashboard for the SRE Team.
+- Defining SLIs and SLOs as well as proposing template SLAs for a managed Keycloak cluster.
+- Defining metrics required by the proposed dashboard as well as designing a mechanism for exporting them to prometheus.
+- Design logging improvements for better error tracking.
+
+What's out of the scope:
+- Traceability - tracking a user across the whole authorization flow. At the moment we don't have any requirements
+in this direction. If we get such a requirement, we will track this progress in a separate epic and design.
```
stianst, line 49: We do have "requirements in this direction", so just write that it is out of scope at this stage, and will be covered in a separate design/epic in the future

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+
+![SRE Dashboard](img/observability-dashboard.png)
+
+NOTE: The diagram implementation and specific charts might change during the implementation phase. The above proposal
+shows only the general idea.
+
+The proposal consists of 3 sections:
+- System health - Enables the SRE Team to quickly tell if Keycloak cluster is up and running. This section consists of:
+  - Readiness and Liveness Probes chart
+  - HTTP 5xx error rate for all major Keycloak Services, including: Admin Console, Account Console, OIDC, SAML
```
stianst, line 61: We probably want a separate metrics for authentication, where "authentication" is the "optional" bit in the middle of an OIDC or SAML authentication request.

There's other services like authz services, token exchange that we probably want to cover as well.

Should probably just be admin or account, not admin console and account console, as the "console" part is not all that interesting since it's just static files.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+  - Aggregated latency across the most important endpoints
```
stianst, line 62: Not sure this has that much value, especially since password hashing is expensive.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+  - (Optional) Error rate in logs [^1]
+- Detailed Metrics - This section helps foresee upcoming failures. It includes:
+  - Pod memory, CPU and container storage charts
+  - GC Times
+  - Heap and non-heap memory metrics
```
stianst, line 67: Detailed Metrics - What about system resources, or something like that instead?

Can we think about a way to make metrics understandable to a non-Java guy?

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+- Experimental metrics - Acts as a bridge between the SRE and Keycloak Teams. We introduce new metrics here and if they
+are useful, we promote them to other sections. This section includes:
+  - External IdP state (Social Login Providers)
+  - The Total number of CRs discovered by the Operator vs the number of CRs successfully reconciled
```
stianst, line 71: I don't like the idea of experimental metrics. It's just a dumping ground for things that we think may or may not make sense.

I'd say we at least need one more section which is "external services" where external IdPs, user fed, db health, etc would be shown.

stianst: Total number of CRs discovered by the Operator - this cries to me that there maybe should be a separate dashboard and metrics for the Operator.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+(see [here](https://github.com/keycloak/keycloak/blob/cee884e4a72c08b7ecafd965ab6f8c430b17ef15/distribution/feature-packs/server-feature-pack/src/main/resources/modules/system/layers/keycloak/org/keycloak/keycloak-server-subsystem/main/server-war/WEB-INF/jboss-deployment-structure.xml)).
+
+The Quarkus-based distribution (Keycloak.x) requires adding `smallrye-metrics` extension (see [the manual](https://quarkus.io/guides/microprofile-metrics)).
+Once this is done, the metrics will be automatically available in `/metrics` endpoint. From the implementation perspective,
+we need to perform the same steps as we were on Wildfly.
+
+## Metrics for the REST interface
+
+Most of the metrics are related to the various REST endpoints. Therefore, it makes sense to put the root metrics
+component into `services` module under its own package - `org.keycloak.services.metrics`.
```
stianst, line 127: We may want to introduce our own abstraction on top of smallrye/microprofile metrics. Also, we probably want to have the ability to plug-in custom metrics providers into Keycloak.

stianst: Further, we should have 1 health endpoint, and 1 metrics endpoint. This should be covered in this design.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+The last part is tracking external IdP responses. Implementing this part requires changing the existing code around
+`SocialIdentityProvider`s and `OIDCIdentityProvider`. All `SocialIdentityProvider` implementations usually throw
+`IdentityBrokerException`s. The exception needs to be enhanced to contain an information about that provider has thrown
+it. Then, the implementation should involve updating all places in `AbstractOAuth2IdentityProvider` and add
+proper checks in all `catch` clauses. Once an `IdentityBrokerException` is caught, and it contains a provider ID,
+it should result in updating certain metrics in the metrics components after logging an error.
+
+## Logging architecture
+
+Collecting logs in an OpenShift cluster is a cross-cutting concern that needs a unified solution. OpenShift provides a
```
stianst, line 159: Is this just a log4j logger, or something else?

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+teams are likely to introduce [Loki](https://grafana.com/oss/loki/) - a Grafana add-on for viewing
+logs. So, before implementing this piece of functionality - please consult it with MAS and AOS Logging Teams.
+
+Depending on timing and other teams progress, it is possible to implement error rate tracking using metrics. This
+however, introduces a few design flaws:
+- A new `LoggerAdapter` would need to be created. Whenever an error is logged in Keycloak's codebase, the `LoggerAdapter`
+will increase the error log counter.
+- All `Logger#getLogger` calls would need to be replaced to acquire the `LoggerAdapter` instead of pure
+JBoss Logging `Logger`.
+- The error-rate metric will miss all error emitted by Wildfly (as they will be thrown in different subsystems).
```
stianst, line 181: Not a chance in hell we're doing this ;)

Log4j (and probably JBoss logging) allows intercepting log events though, so you don't need to have a different logger to handle that.

stianst (commented):
As part of the design of observerability I want a list of failure scenarios, and a report analysing how the proposed design would allow SRE staff to detect and resolve the issues covered in the failure scenarios. That is the only way to verify what we are building is what is needed.

Another thing that is missing in this design is more details around logging. What logging do we have that is useful if any? Is there logging that is missing (related to failure scenarios)? What about log entry numbers?

Also, what about Keycloak events? I can see they are collecting a lot of useful information, but this is not covered at all in this design.

stianst:
This is a great start, but there's still quite a bit that needs to be done before we can agree on what exactly it should look like. One very important thing is listing failure scenarios, and how dashboard, logging, etc. would help SRE staff to identify and resolve those scenarios. Without it I feel we are kinda shooting in the dark.

design/observerability-2.md
bs-matil: For us it would also be interesting to combine this view with the kubernetes statistics. So one would have the internal and the external view on the workload. For example if applied resource limits for cpu and memory could be made visible within the same view

design/observerability-2.md
slaskawi: The resources you mentioned were the reason I decided to go for a separate document. Most of the are non-public and out-of-date but nevertheless they might be useful in some contexts. Once I replace the document, we'll probably lose them (I can't imagine someone searching through our git repo to dig them out). 

So let me ask you one more time - shall I replace that document? Or maybe I convinced you and we can leave it as is - as a separate doc?

design/observerability-2.md
slaskawi: Ok, will do that. I'll also find a Tractability epic JIRA and link it here.

design/observerability-2.md
slaskawi: Ok, so let's try to identify the course-grained services we want to expose:
- Admin
- Account
- OIDC
- Token Exchange
- SAML

Token Exchange definitely doesn't belong to "OIDC" as it's part of the oAuth spec. So it needs to be separate.

SAML and authentication has a lot in common conceptually but they use different endpoints. So I think they can be separate. 

Does this split works for you?

design/observerability-2.md
slaskawi: I totally agree. However with such an aggregate (average latency over multiple endpoints in time) you can easily spot the trend. Once it goes up, you know you might be in troubles pretty soon. I believe it's super valuable.

design/observerability-2.md
slaskawi: @stianst The purpose of the "Detailed Metrics" is to provide the SRE Team more detailed information. This is very useful when reporting JIRAs against our team. We need to go a bit further than typical OS metrics here to make such reports more valuable. Keycloak is a Java-based app. The SRE Team needs to have some basic training in monitoring JVM-based apps. I don't see other way.

@bs-atil Putting Resources and Limits in the same graph seems like a good idea. Note, that Pod Memory and CPU metrics are reported by OpenShift/Kubernetes directly from CGroups. Those are the once we are interested in on left side of the dashboard (Pod Memory, CPU and Storage). The right side is more JVM-related. There you have Heap/non-heap and GC times. Perhaps it makes sense to put Kubernetes Resources in Heap/non-heap graph as well. I'll try to incorporate this into the sketch diagram and see how it looks like.

design/observerability-2.md
slaskawi: > I don't like the idea of experimental metrics. It's just a dumping ground for things that we think may or may not make sense.

I believe we need some section that we'll use to introduce new metrics. They might misbehave at first and I wouldn't like anyone to be page based on them. If they live in an area called "Experimental" - the contract seems to be clear. Perhaps you have some suggestion how to improve this?

> Total number of CRs discovered by the Operator - this cries to me that there maybe should be a separate dashboard and metrics for the Operator.

That's actually a good point. There's one more metric we could put there - the liveness/readiness probe. Do you have any other ideas?

design/observerability-2.md
slaskawi: > We may want to introduce our own abstraction on top of smallrye/microprofile metrics. 

I don't like this. We have no idea if Micrometer will implement Microprofile Metrics or not. If it will, implementing our own metrics stuff on top of Microprofile Metrics (instead of calling Smallrye directly) would be enough. Let's keep it simple and follow [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) rule. 

> Also, we probably want to have the ability to plug-in custom metrics providers into Keycloak.

Could you please elaborate? What Metrics Providers do you have in mind? If you're thinking about Micrometer - it depends what the future will bring us. Either Micrometer will support Microprofile Metrics and there will be little-to-none modifications required on our side (we'll just swap the dependencies). Or we'll need adapt to Micrometer by rewriting our code. Unfortunately, it's very early stage of this discussions (they mainly happen within Quarkus Team) and we know nothing at this point. 

> Further, we should have 1 health endpoint, and 1 metrics endpoint. This should be covered in this design.

Yes, I'll put that into the design.

design/observerability-2.md
slaskawi: Keycloak uses JBoss Logging as a logging facade. So this code needs to be applied on top of JBoss Logging.

design/observerability-2.md
slaskawi: I agree. Catching Wildfly error logs is very tricky as the implementation would need to go into Wildfly's Logging Subsystem. I already clarified this with JBoss Logging Team.

slaskawi:
> This is a great start, but there's still quite a bit that needs to be done before we can agree on what exactly it should look like. One very important thing is listing failure scenarios, and how dashboard, logging, etc. would help SRE staff to identify and resolve those scenarios. Without it I feel we are kinda shooting in the dark.

> As part of the design of observerability I want a list of failure scenarios, and a report analysing how the proposed design would allow SRE staff to detect and resolve the issues covered in the failure scenarios. That is the only way to verify what we are building is what is needed.

@stianst I agree. Adding the failure scenarios and showing how to resolve them is a good idea. I'll add this in the next iteration.

> Another thing that is missing in this design is more details around logging. What logging do we have that is useful if any? Is there logging that is missing (related to failure scenarios)? What about log entry numbers?

@stianst I had some discussions around this with @david-martin. The main conclusion was that log messages are quite useful when investigating failures. Detecting whether a service is up and healthy is should happen based on metrics. So in this sense, log messages are a secondary priority to this design.

Perhaps, it would satisfy your requirements if I showed the number of error logs emitted per each failure scenario? This way we would have at least some idea about the logging but I wouldn't need to go into the details. If this doesn't work for you - please suggest how would you like me to explore this area.

> Also, what about Keycloak events? I can see they are collecting a lot of useful information, but this is not covered at all in this design.

@stianst After having a conversation with @david-martin, @maleck13 and @pb82 it seems the SRE Team is not interested (at least in the first iteration of this feature) in things like the number of successful/failure logins etc. Those metrics were not included in the SRE Dashboard, thus there's no need to expose them. In other words once this feature gets implemented and the Keycloak Operator will produce an updated SRE Dashboard, we'll drop the Aerogear extension (if that's what you're asking between the lines). Perhaps this conversation will be resumed when considering Observability v2 in the future.  

@david-martin, @maleck13 and @pb82 Please let me know if I didn't interpret our previous conversations correctly and this is something we need.

> I really like this new proposal.

@bs-matil Thank you!

> We separate the view on the API levels.
> so -> how healthy is the account API vs how health the admin API vs the authentication/authorization APIs.

@bs-matil We'd like to divide Keycloak into a course-grained services, like Account, Admin, OIDC etc. (see the conversion in one of the comments). Then, we'll track the health per each of those course-grained components. 

> Why? We judge different API's less critical to the customer. E.g. if the admin API fails it mostly does not affect the end-user while an issue in the login flows and authorization endpoints will affect the whole userbase.

@bs-matil Individual endpoints might not be the best way to do it. Imagine oAuth Authorization endpoint being up (`<cut-url>/auth') but for some reason the Token endpoint being down ('<cut-url>/token'). In this case OIDC (technically oAuth, but I believe we'll put this in OIDC component) would report being down. I believe that's a much more valuable information than the Token endpoint is down and Authorization endpoint being up. The end result is that oAuth/OIDC service doesn't work.

> Does this make sense? A simple view on this would be additonal series displayed in the error graphs by api

@bs-matil We prefer to track it via Keycloak Service. See my explanation the above.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+
+Based on metrics, dashboards and alerts, it is possible to define SLA - an agreement between a Service Provider and a
+Service Consumer.
+
+## Scope
+
+What's in the scope of this design:
+- Designing a new Healcheck API. This has been covered in a separate design document. See [KEYCLOAK-12398](https://issues.redhat.com/browse/KEYCLOAK-12398)
+- Designing and implementing a dashboard for the SRE Team.
+- Defining SLIs and SLOs as well as proposing template SLAs for a managed Keycloak cluster.
```
david-martin, line 43: Any SLOs defined here would be 'possible SLOs' or 'suggested SLOs'.
Agreeing on objectives is a collaborative process between the different stakeholders (i.e. who's running the service (SRE) and what the business wants.

re: SLAs, I would suggest steering clear of any recommendations or proposals here and stick to just SLOs.
It could get hairy given that an SLA is a legal document.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+- Defining metrics required by the proposed dashboard as well as designing a mechanism for exporting them to prometheus.
+- Design logging improvements for better error tracking.
```
david-martin, line 45: This is good to have, but a little fuzzy as to what it means.
Is there something specific around logging you're thinking of here? Something SRE use case maybe?

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+|------------------------|--------------|-------------------------------------------------------------------------------|---------------|----------------------------------------------------------|
+| Health                 | Availability | Check individual Pod Liveness and Readiness status                            | To be defined | Check the DB status  / Report a JIRA / Restart Pods      |
+| Health                 | Error rate   | Monitor error rates of individual Keycloak services                           | To be defined | Investigate further                                      |
+| Health (optional)      | Error rate   | Monitor error rate in logs                                                    | To be defined | If increases, check what might be causing it             |
+| Performance            | Latency      | Monitor the aggregated average response time for individual Keycloak Services | To be defined | If drastically increases, check what might be causing it |
+
+For all the SLOs mentioned the above, the Keycloak Operator needs to set proper Prometheus Alerts.
```
david-martin, line 95: This is useful.
Is there a plan to document the `SRE Team Action if failing` in more detail? i.e a SOP or runbook

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+- A new `LoggerAdapter` would need to be created. Whenever an error is logged in Keycloak's codebase, the `LoggerAdapter`
+will increase the error log counter.
+- All `Logger#getLogger` calls would need to be replaced to acquire the `LoggerAdapter` instead of pure
+JBoss Logging `Logger`.
+- The error-rate metric will miss all error emitted by Wildfly (as they will be thrown in different subsystems).
+
+Keycloak.x distribution offers logging the number of log events by enabling `quarkus.log.metrics.enabled`. The functionality
+has been implemented in Quarkus using [this Pull Request](https://github.com/quarkusio/quarkus/pull/11224).
+
+Having the above in mind, if the implementation choice is to expose the number error in logs using metrics endpoint,
```
david-martin, line 186: Is it possible to expose metrics for the underlying causes of the errors?

design/observerability-2.md
abstractj: @slaskawi I don't think we will lose them, as long as we don't force push. There's always git history for it.

abstractj: @stianst @slaskawi would make sense to get some inspiration from Kiali? https://kiali.io/documentation/latest/validations/. Just as an example: https://kiali.io/documentation/latest/validations/#_kia0002_more_than_one_selector_less_object_in_the_same_namespace

design/observerability-2.md
abstractj: @slaskawi should we mention User Federation? I'm thinking about how people could identify an issue when users struggle to login with their LDAP credentials for example.

design/observerability-2.md
slaskawi: @abstractj Could you please put specific examples inline to this design? The document you linked is pretty big and touches many different things. It's hard for me distill what inspiration you're talking about.

slaskawi: I also replaced the `observability.md` document.

design/observerability-2.md
slaskawi: Updated.

design/observerability-2.md
slaskawi: @abstractj Yes, that works for me.

design/observerability-2.md
slaskawi: Added both Resources and Limits to the graphs.

design/observerability-2.md
slaskawi: Introduced another category - "Operator Metrics".

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+- EFK Stack - Elsticsearch, Fluentd and Kibana. A typical log-collecting stack.
+- CLO - Cluster Logging Operator
+- CR - Custom Resource. An API provided by Keycloak Operator for the context of this document.
+
+## Motivation
+
+Managing Keycloak cluster requires involving a dedicated SRE (Site Reliability Engineering) Team. Such a team needs to
+be able to:
+
+* Diagnose if Keycloak cluster is up and running
```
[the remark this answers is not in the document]
slaskawi: Totally agree. Let me add this to the design.

design/observerability-2.md
```diff
@@ -0,0 +1,196 @@
...
+Typically, an SRE Team uses dashboards and alerts. Both of them are created based on metrics exported by a Service
+(in this case, the Service is both Keycloak and Keycloak Operator).
+
+Based on metrics, dashboards and alerts, it is possible to define SLA - an agreement between a Service Provider and a
+Service Consumer.
+
+## Scope
+
+What's in the scope of this design:
+- Designing a new Healcheck API. This has been covered in a separate design document. See [KEYCLOAK-12398](https://issues.redhat.com/browse/KEYCLOAK-12398)
```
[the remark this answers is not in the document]
slaskawi: Fixed. Thanks!

design/observerability-2.md
slaskawi: That's a very good point! Let me clarify this in the design.

design/observerability-2.md
slaskawi: Added a new "Logging Improvements" paragraph and put some details around the implementation.

design/observerability-2.md
slaskawi: With the logging improvements I mentioned in the new version of the design - it should be possible to do that. Loki should enable you to select logs with a specific phrase, like "KC-SERVICES0002". If we used the prefix consistently, you should be able to build such metrics ad-hoc.

design/observerability-2.md
slaskawi: I think such a document should be written as part of the documentation. Let me mention this explicitly.

design/observerability-2.md
slaskawi: I believe this one has been resolved. Please reopen if you think otherwise.

design/observerability-2.md
slaskawi: Explained. Resolving.

design/observerability-2.md
slaskawi: Resolving as explained.

slaskawi:
Dear all,

I've just uploaded the third iteration of the Healthcheck proposal.

Differences from the previous version:
* Proposed the new split for Keycloak Services: Admin, Account, OIDC, SAML, Token Exchange and User Federation
* Updated the SRE Dashboard:
  * Included the new "Session Counts" in the "Detailed metrics" section. There was an RFE for this
  * Introduced "Operator metrics"
  * Added "User Federation" in "Experimental metrics"
* Added the "Investigating failures" paragraph
* Added more information about Logging Improvements (now in a separate paragraph)
* Switched the implementation for Micrometer

Feedback is more than welcome!

Thanks,
Sebastian

design/observerability.md
```diff
@@ -1,150 +1,263 @@
...
-and will be considered in the future. 
+Both implementations (Wildfly and Quarkus based) will use a similar process for implementing metrics. The Micrometer
+has a notion of a [Gloabl Registry](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/Metrics.java#L35)
+that can be used for environments where CDI injection is not possible (as a reminder, Keycloak doesn't use CDI). The implementation
+can obtain all metrics instruments (such as gauges, histograms etc.) from it.
 
+## Metrics for the REST interface
 
-## Resources
+Most of the metrics are related to the various REST endpoints. Therefore, it makes sense to put the root metrics
```
kenfinnigan, line 193: Depending on what library you use for REST endpoints, there might already be a Micrometer Binder that's applicable to automatically provide those metrics.

If not, it might be possible to utilize what has been done for Quarkus, if you're using Vert.x underneath.

As a last resort, developing a `MeterBinder` directly shouldn't be too hard

abstractj:
@slaskawi would be possible to add how we plan to protect those endpoints? I can think about mTLS, but it would be nice to make it clear into the document.

Asking because this may have an impact in the Healcheck proposal https://github.com/keycloak/keycloak-community/pull/55, and also in the implementation details.

design/observerability-2.md
abstractj: @slaskawi my suggestion was pretty much regards the structure followed by Kiali, to provide scenarios and how to solve each issue. But I guess you kind of did it here https://github.com/keycloak/keycloak-community/blob/af2de8a2c4ed2cc0b12e3e789379239c8bdba1bd/design/observerability.md#investigating-failures

pedroigor:
@slaskawi I also think that Keycloak Events are an important source of information and very useful when monitoring the environment. It is not about the number of (un)successful login attempts only but monitor the system usage and detect possible anomalies. I can give more examples of how they are important, but even a simple `EMAIL_SEND_FAILED` may indicate critical problems that may be impacting the business and the overall availability of the deployment.

I also missing information about:

* Response times. Mainly for authentication and/or any other service that impacts end-user. It should help to monitor usage and track anomalies once a baseline is created based on history. 
* Datasource
* Thread Pool
* Downstream Services which the server relies on. For instance, SMTP Server, LDAP Server, etc (Datasource too, as mentioned above)
* Session Counts. In addition to what you have, I would say that some other metrics are important like those related to eviction and removal. Possibly, separate graphs for the most critical caches.
* Cluster. Not sure if this is possible, but somehow monitor cluster partitions. This is usually related to unavailability issues and may indicate problems in other areas.

I also think that `Coarse-grained Keycloak Services` are too coarse-grained. For instance, what OIDC really means and what the Aggregated Latency Graph is really showing? Is it related to authorization requests, or token requests, or refresh token requests? The same goes for SAML, is it related to authentication requests or responses?

slaskawi:
> would be possible to add how we plan to protect those endpoints? I can think about mTLS, but it would be nice to make it clear into the document.

@abstractj That's a good point. I'll include this in the next iteration

> I also think that Keycloak Events are an important source of information and very useful when monitoring the environment. It is not about the number of (un)successful login attempts only but monitor the system usage and detect possible anomalies. I can give more examples of how they are important, but even a simple EMAIL_SEND_FAILED may indicate critical problems that may be impacting the business and the overall availability of the deployment.

@pedroigor I totally agree. However, as I outlined in [my previous comment](https://github.com/keycloak/keycloak-community/pull/177#issuecomment-698915272) the SRE Team is mostly interested in keeping Keycloak alive. Checking if Keycloak is under attack and other similar things (@stianst called it "Security as a Service" at some point - which is a quite good term) is definitely something we should think for the future.

I will create a separate paragraph, called "Future Work" and collect all those ideas in a single place. Does it work for you?

> I also think that Coarse-grained Keycloak Services are too coarse-grained

@pedroigor This has been agreed (and actually proposed by) @stianst. Perhaps you two could do a small discussion and come up with a single proposal. I'm open for suggestions here.

> For instance, what OIDC really means and what the Aggregated Latency Graph is really showing

@pedroigor This would be an aggregate latency (a histogram to be precise) of the response times over authorization and token endpoint. This is something to start with. The main idea is to get some average latency and observe if it goes up very quickly. Such a behavior might indicate performance problems coming your way.

slaskawi:
Today morning, both @stianst and I had a discussion on the future plans of both Healthcheck API and the Observability designs. 

We agreed that our main focus will be Quarkus-based distribution. As for the time being, we'll use what we have for the Wildfly-based distribution (which translates to the [embedded metrics](https://github.com/keycloak/keycloak-containers/tree/master/server#enable-some-metrics), custom [Aerogear Extension](https://github.com/aerogear/keycloak-metrics-spi) for additional metrics and the [Healthcheck](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_probes.go) as they are now). Both Designs will be repurposed and updated in the following days.

One of the crucial differences is that we agreed to verify if we can use CDI extension in Quarkus (just to provide additional context - the CDI has been removed from Wildfly-based distribution as it introduces additional latency and increased the bootstrap time). Both Microprofile Health as well as Micrometer implementation provide CDI extensions to easily wire everything together. 

//cc @stianst @ASzc @miquelsi @abstractj @vmuzikar @pedroigor

pedroigor:
> @pedroigor I totally agree. However, as I outlined in [my previous comment](https://github.com/keycloak/keycloak-community/pull/177#issuecomment-698915272) the SRE Team is mostly interested in keeping Keycloak alive. Checking if Keycloak is under attack and other similar things (@stianst called it "Security as a Service" at some point - which is a quite good term) is definitely something we should think for the future.

Now I see your comment. Sorry. But still, events should give useful information about attacks. But I understand it is out of scope for now.

> 
> I will create a separate paragraph, called "Future Work" and collect all those ideas in a single place. Does it work for you?

Don't worry if it does not make sense to include this document.

> 
> > I also think that Coarse-grained Keycloak Services are too coarse-grained
> 
> @pedroigor This has been agreed (and actually proposed by) @stianst. Perhaps you two could do a small discussion and come up with a single proposal. I'm open for suggestions here.

Take OIDC for instance. IMO, both authorization and token endpoint should be considered a service. These are the main on involved in the different flows when users are authenticating. Depending on how people would use us, I would also have both token introspection and userinfo separated. Grouping all those in a single adds little value because it won't be easy to visualize what is going on.

But again, I'm late here :) So maybe this another thing that we'll looking later?

But yeah, other things that I mentioned such as the Thread and Database Pools, as well as caches, are quite crucial, IMO.

Maybe, SRE should talk with some people using Keycloak and Grafana (we can talk about which ones in private as they are customers) to get some inspiration.

> 
> > For instance, what OIDC really means and what the Aggregated Latency Graph is really showing
> 
> @pedroigor This would be an aggregate latency (a histogram to be precise) of the response times over authorization and token endpoint. This is something to start with. The main idea is to get some average latency and observe if it goes up very quickly. Such a behavior might indicate performance problems coming your way.

Yeah, but still too high level. But it is OK if SRE is fine with it for now.

design/observerability.md
```diff
@@ -1,150 +1,263 @@
...
+The proposal consists of 3 sections:
+- System health - Enables the SRE Team to quickly tell if Keycloak cluster is up and running. This section consists of:
+  - Readiness and Liveness Probes chart
+  - HTTP 5xx error rate for all major Keycloak Services, including: Admin Console, Account Console, OIDC, SAML
+  - Aggregated latency across the most important endpoints
+  - (Optional) Error rate in logs [^2]
+- Detailed Metrics - This section helps foresee upcoming failures. It includes:
+  - Pod memory, CPU and container storage charts
+  - GC Times
+  - Heap and non-heap memory metrics
```
hmlnarik, line 75: Per-cache utilization should be here as a separate item for the time being to display whether configuration needs tweaking.

hmlnarik (commented):
Thank you @slaskawi, the proposal looks good to me!

Similarly to #55, I would like to see an example of a successful and unsuccessful response in the design. Could it be added?

Albeit I understand this is currently out of scope for reasons stated above, I completely agree with @pedroigor that events are important source of attack detection and should eventually find the way to the observability endpoint, at least their count per event type. Implementation of these should be pretty straightforward once the observability endpoint is up and running.

Also, would there be a single observability endpoint or would the metrics be somehow split into categories? From performance perspective, it makes sense the latter. The spec seems to leverage tags for this purpose. If we move this direction, I believe the tags or any other means of metric separation should be part of this proposal.

design/observerability.md
slaskawi: Please note, that the new Observability initiative targets Quarkus deployment, which by definition will use a remote Infinispan cluster. Having this in mind, I'm not sure if monitoring Infinispan caches should be our concern. Perhaps this is something for the Infinispan Observablity? @ryanemerson - do you have any thoughts around this?

design/observerability.md
```diff
@@ -1,150 +1,263 @@
...
+  - The number of Infinispan sessions
```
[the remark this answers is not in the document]
slaskawi: The implementation might be chosen a bit later but I wanted this to be a number of session seen by Keycloak. The number of entries in Infinispan caches might be an implementation detail here.

slaskawi:
> Similarly to #55, I would like to see an example of a successful and unsuccessful response in the design. Could it be added?

@hmlnarik In case of metrics, we don't take a failure scenarios into account. The fact that Prometheus can not scrape metrics (e.g. the `/metrics` endpoint returns HTTP 500) needs to be reported at Platform Health level. I imagine there are separate dashboards to do that.

> Albeit I understand this is currently out of scope for reasons stated above, I completely agree with @pedroigor that events are important source of attack detection and should eventually find the way to the observability endpoint, at least their count per event type. Implementation of these should be pretty straightforward once the observability endpoint is up and running.

> Now I see your comment. Sorry. But still, events should give useful information about attacks. But I understand it is out of scope for now.

@pedroigor @hmlnarik Ok, so what type of events would you like to see on the dashboard and why? Perhaps for the first iteration we should just have a histogram of what events were emitted? It's generic enough and let's you build other, more complicated dashboard based on this information. 

> Also, would there be a single observability endpoint or would the metrics be somehow split into categories? From performance perspective, it makes sense the latter. The spec seems to leverage tags for this purpose. If we move this direction, I believe the tags or any other means of metric separation should be part of this proposal.

@hmlnarik No, there's only one `/metrics` endpoint, which is scaped by Prometheus. 

> Take OIDC for instance. IMO, both authorization and token endpoint should be considered a service. These are the main on involved in the different flows when users are authenticating. Depending on how people would use us, I would also have both token introspection and userinfo separated. Grouping all those in a single adds little value because it won't be easy to visualize what is going on.

@pedroigor I think putting both of them into the same category makes more sense. Imagine an authorization endpoint being down and token endpoint being up. What does it mean? Some parts of the OIDC spec might work and some might not. But from the user perspective - it's mostly down (you can not log in, right). This is why I think it makes more sense to put them into a single category. 

> But yeah, other things that I mentioned such as the Thread and Database Pools, as well as caches, are quite crucial, IMO.

@pedroigor I wouldn't go that far to be honest. At least not for an initial implementation. To me it's too fine-grained.

PR #181, opened 2020-10-12
KEYCLOAK-7301 No-downtime-capable storage

hmlnarik:


stianst (commented):
LGTM

slaskawi (approved):
LGTM - an excellent design, @hmlnarik !

design/keycloak.x/storage-persistence.md
```diff
@@ -0,0 +1,530 @@
...
+that is to ensure that every store version _N + 1_ writes two object
+versions - one for version _N_ and other _N + 1_. There are other ways
+possible to achieve the same, basically they count on redundancy.
+
+### Physical Schema Upgrades
+
+Some of the changes need updates of the physical storage schema. For example,
+in case of adding a single searchable field, this field should be somehow
+indexed, and in case of relational databases, that may require adding
+a column and an index. Adding an index may in turn lock the table and
```
mposolda, line 458: So does this mean that new store may still somehow support storing fields in some RDBMS like format where the indexed field will require to have it's separate columns?

If that is the case, I wonder if it would makes sense to store indexes in separate tables? For example when there is an existing table ROLE and then there will be a need to create the index on field "name", there will be a need to create new table ROLE_NAME_IDX, which will contain just the ID of the role and the NAME of it. This will mean that existing table ROLE will be still able to "serve requests" from the old version and users in the old version won't be "blocked" by adding new index?

design/keycloak.x/storage-persistence.md
```diff
@@ -0,0 +1,530 @@
...
+
+It is obvious that no-downtime requirement and performance are mutually
+outbalancing each other. Hence it should be stated that:
+
+* No-downtime constraints must be satisfied
+* Performance needs to be good-enough. This means that:
+  * Object lookup by an object identifier must succeed
+    in logarithmic time at worst
+  * Lookup by searchable attribute should be bound
+    by logarithmic time
```
mposolda, line 197: I suppose the "logarithmic time" is meant to be compared to the total count of existing objects in the store, right? Probably just a nitpick, but it may not be 100% obvious from the text IMO :-)

design/keycloak.x/storage-persistence.md
```diff
@@ -0,0 +1,530 @@
...
+
+* _Keep the store schema unchanged._ Update the nodes gradually
+  to the new version.
+
+  With the current database (as of Keycloak 11), this would work
+  for most of the maintenance releases. However it would
+  not work if there were new features that introduce more
+  than cosmetic changes the database.
+
+The last option is attractive though. Could the store structural changes
```
mposolda, line 60: Probably a nitpick, but it may not be 100% clear what exactly means "structural" changes and what is "change in object models" ? The latter is described below though, but at the point when newbie person is reading this, it may be a bit confused. But maybe it is just me :-)

mposolda (commented):
@hmlnarik Thanks for the design, looks good! Just added some minor comments below.

design/keycloak.x/storage-persistence.md
hmlnarik: > So does this mean that new store may still somehow support storing fields in some RDBMS like format where the indexed field will require to have it's separate columns?

Yes, actual layout depends on the implementor.

> If that is the case, I wonder if it would makes sense to store indexes in separate tables? For example when there is an existing table ROLE and then there will be a need to create the index on field "name", there will be a need to create new table ROLE_NAME_IDX, which will contain just the ID of the role and the NAME of it. This will mean that existing table ROLE will be still able to "serve requests" from the old version and users in the old version won't be "blocked" by adding new index?

Yes, this may be an option. Yet another option may be to leverage database capable of building indices on JSON expressions, e.g. https://bitnine.net/blog-postgresql/postgresql-internals-jsonb-type-and-its-indexes/

design/keycloak.x/storage-persistence.md
hmlnarik: Yup, will update the design

design/keycloak.x/storage-persistence.md
mposolda: Cool, I wasn't aware of this option. I am sure that you will eventually figure the most optimal way for each RDBMS :-) Just wanted to clearify this a bit...

design/keycloak.x/storage-persistence.md
hmlnarik: I rephrased it in the new version. Thank you for the review!

mposolda (approved):
@hmlnarik Thanks!

stianst (approved):

design/keycloak.x/storage-persistence.md
```diff
@@ -0,0 +1,532 @@
...
+For example, in Keycloak of version 1.2.3, versions can be 53 for user objects,
+41 for clients, and 14 for realms. Versioning of Keycloak and the logical objects
+is independent, so it is possible that the Keycloak version 1.2.4 may have
+the same object versions as in 1.2.3, that one of them has changed,
+or all have changed.
+
+*Note on testability:* It must be possible to automatically test that
+a store implementation satisfies all the requirements stated in this section.
+For this purpose, there is
+[a project in keycloak-playground](https://github.com/keycloak/keycloak-playground/no-downtime-upgrade/) repository
```
lhanusov, line 169: @hmlnarik Please, replace current playground url with the valid one: https://github.com/keycloak/keycloak-playground/tree/master/no-downtime-upgrade

hmlnarik:
Fixed broken links and a method name. Merging as the design proposal has been implemented in a map storage in the main Keycloak tree, and there will be follow-up specs for individual storages.

PR #188, opened 2020-11-09
KEYCLOAK-16137 Client Policy : Support New Admin REST API

tnorimat:
This PR is for [KEYCLOAK-16137 Client Policy : Support New Admin REST API](https://issues.redhat.com/browse/KEYCLOAK-16137).  It is part of the work for realizing Client Policy Official Support as [FAPI-SIG activity](https://github.com/keycloak/kc-sig-fapi/projects).

This PR adds description of new Admin REST API for Client Policies to the existing Client Policies' design document.

### DIFF
--- a/design/client-policies.md
+++ b/design/client-policies.md
@@ -1,8 +1,7 @@
 # Client Policies
 
-* **Status**: Draft #1
-* **JIRA**: [KEYCLOAK-13933](https://issues.jboss.org/browse/KEYCLOAK-13933)
-
+* **Status**: Draft #2
+* **JIRA**: [KEYCLOAK-13933](https://issues.jboss.org/browse/KEYCLOAK-13933), [KEYCLOAK-16137](https://issues.jboss.org/browse/KEYCLOAK-16137)
 ## Motivation
 
 To make it easy to secure client applications, it is beneficial to realize the following four points in a unified way.
@@ -22,7 +21,7 @@ To realize these four points in a unified way, here "Client Policies" concept is
 
 ## Use-cases
 
-Client Policies realize the following three points mentioned in Motivataion as follows.
+Client Policies realize the following three points mentioned in Motivation as follows.
 
 ### 1. Setting policies on what configuration a client can have
 
@@ -40,7 +39,7 @@ The current keycloak has already supported validation on whether the client foll
 
 These can be specified by each setting item (on Admin Console, switch, pulldown menu, etc). To make the client application secure, user need to set a lot of such settings in appropriate way, which makes it difficult for the user to secure their client application.
 
-Client Policies can do these validation of client configurations mentioned just above. In the futre, these each setting items will be replaced by Client Policies. 
+Client Policies can do these validation of client configurations mentioned just above. In the future, these each setting items will be replaced by Client Policies. 
 
 The pre-set client policies for following some security profiles (FAPI, SPA, Native App, etc) are provided, which makes it easy for the user to secure their client application.
 
@@ -55,11 +54,13 @@ This client policy concept is independent of any specific protocol. However, thi
 
 ## Implementation details
 
-Client Policies consists of the four building blocks, Policy, Condition, and Executor.
+Client Policies consists of the four building blocks, Condition, Executor, Profile and Policy.
 
 ### Condition : which client
 
-A condition determins to which client a policy is adopted. The condition checks whether one specified criteria is satisfied. Such the task is that checking whether the [access type](https://www.keycloak.org/docs/latest/server_admin/index.html#oidc-clients) of the client is confidential.
+A condition determines to which client a policy is adopted. The condition checks whether one specified criteria is satisfied. For example, some condition checks whether the [access type](https://www.keycloak.org/docs/latest/server_admin/index.html#oidc-clients) of the client is confidential.
+
+The client can not be used solely by itself. It can be used in a policy that is described afterwards.
 
 #### How to implement
 
@@ -67,7 +68,7 @@ A condition can be implemented as a provider (`ClientPolicyConditionSpi`, `Clien
 
 #### How to configure
 
-A condition can be configuable the same as other configurable provider. What can be configured depends on each condition's nature.
+A condition can be configurable the same as other configurable provider. What can be configured depends on each condition's nature.
 
 #### Which kind of conditions are provided
 
@@ -109,7 +110,9 @@ E.g. When a client sends an authorization request, a policy is adopted if this c
 
 ### Executor : what action
 
-An executor specifies what action is executed on a client to which a policy is adopted. The executor executes one or several specified actions. Such the action is checking whether the value of the parameter "redirect_uri" in the authorization request matches exactly with one of the pre-registered redirect URIs on Authorization Endpoint and rejecting this request if not. 
+An executor specifies what action is executed on a client to which a policy is adopted. The executor executes one or several specified actions. For example, some executor checks whether the value of the parameter "redirect_uri" in the authorization request matches exactly with one of the pre-registered redirect URIs on Authorization Endpoint and rejects this request if not. 
+
+The executor can not be used solely by itself. It can be used in a profile that is described afterwards.
 
 #### How to implement
 
@@ -117,11 +120,11 @@ An executor can be implemented as a provider (`ClientPolicyExecutorSpi`, `Client
 
 #### How to configure
 
-An executor can be configuable the same as other configurable provider. What can be configured depends on each executor's nature.
+An executor can be configurable the same as other configurable provider. What can be configured depends on each executor's nature.
 
 #### What events
 
-An executor acts on the follwing events:
+An executor acts on the following events:
 
 * Creating a client
 * Updating a client
@@ -151,34 +154,48 @@ One of several purposes for this executor is to realize client conformance profi
 
 Also, one of several purposes for this executor is to take over [Client Registration Policy](https://www.keycloak.org/docs/latest/securing_apps/index.html#client-registration-policies).
 
+### Profile : bundle executors to realize a security profile
+
+A profile consists of several executors which can realize a security profile like FAPI.
+
+The profile is referred from the policy just described below and all executors in this profile are executed against the clients which this policy adopts. 
 
 ### Policy : how to determine which client and what action
 
-A policy consists of several conditions and executors. A policy can be adpoted to a client satisfying all conditions of this policy. All executors execute their task against this client.
+A policy consists of several conditions. The policy can be adopted to clients satisfying all conditions of this policy. The policy refers several profiles and all executors of these profiles execute their task against the client that this policy is adopted.
+
+This policy can be enabled/disabled. The policy works if it is enabled.
 
 Considering both a condition and an executor do their task independently from other conditions and other executors, the administrator can not specify in which order conditions and executors are evaluated.
 
-#### Pre-set Policy
+#### Built-in Policy and Profile
 
-A pre-set policy is a client policy that have registered in keycloak as default.
-This pre-set policy can be used to enforce basic security level to all clients.
+A built-in policy and profile are ones registered in keycloak as default and can not be modified. This built-in profile can be used to enforce basic security level to all clients.
 
-### Endpoint and Opeation : when and where policy is adopted
+### Endpoint and Operation : when and where policy is adopted
 
 Conditions and executors of a policy should be evaluated on endpoints where a client application interact with keycloak by OAuth2/OIDC protocol. Considering this point, when and where a policy can be adopted to a client is as follows:
 
 ##### OIDC Dynamic Client Registration
-  * Opeation
-    * Register
-    * Update
+  * Operation
+    * Before Register Client
+    * After Registered Client
+    * Before Update Client
+    * After Updated Client
+    * View Client
+    * Unregister Client
   * Authentication
     * Authenticated (Using Initial Access Token, Registration Access Token)
     * Anonymous
 
 ##### Admin REST API
-  * Opeation
-    * Register
-    * Update
+  * Operation
+    * Before Register Client
+    * After Registered Client
+    * Before Update Client
+    * After Updated Client
+    * View Client
+    * Unregister Client
 
 ##### Authorization Endpoint
   * Operation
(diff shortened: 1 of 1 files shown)


tnorimat:
@mposolda This PR is about what I mentioned in [page 15 of FAPI-SIG 6th Meeting Agenda](https://github.com/keycloak/kc-sig-fapi/blob/master/FAPI-SIG/meetings/6th/presentations/FAPI-SIG_6th_MTG_agenda%20.pdf). I'm afraid I'm not sure whether you are an appropriate person to review it. Could you or your team review this PR?

mposolda (changes requested):
@tnorimat Thanks for the design update!

As discussed already, it will be good to have the concept of "Update client policy as JSON" available from the admin console . Also it will be good to have the concept of preSet/builtin client policies, which won't be updateable. It's nice that your design covers this. However I am not sure about some details - especially if we need separate `representation` objects for everything and how exactly to track builtin/preSet components and represent them in the realm JSON export etc. This will be good to clarify with @stianst . Some of my thoughs for this are as follows:

- I think we don't need separate java types like `ClientPolicyConditionRepresentation`, `ClientPolicyExecutorRepresentation` and `ClientPolicyRepresentation`. IMO we can live with the existing type like `ComponentRepresentation` but maybe slightly update it for the case of better support of the preSet/builtin policies?

- I think that none of condition, executors and policies need `subType` field to be available. The policy may need `description`, but IMO conditions and executors does not need `description` on itself?

- I understand that it will be good to have some set of "pre-set" policies (built-in policies), which will be set by default. Few points to this:
    - Field "subType" is probably not the best place to track whether the clientPolicy is built-in or not. IMO this should be tracked elsewhere as the information if policy is builtin policy is clearly not a "subType" .
    - I agree that builtin/pre-set policies should not be creatable/updateable/removable. However IMO we may need at least some way to disable them somehow? WDYT @stianst ?
    - Regarding this, I am thinking if we add the fields like "preSet" and "enabled" on the ComponentModel and ComponentRepresentation (or alternatively create subclasses of those with the new fields).
    - In the realm export JSON files, we don't need the details about the "preSet" policies. However we may just need the info if the policy is enabled or not. For other manually created policies (those which are not preSet), we need to include full info. In the realm export JSON, we may have just some information like this:

```
{      
    "name" : "builtin-policy-xy",
    "preSet" : true,
    "enabled" : true
},
{      
    "name" : "builtin-policy-some-other",
    "preSet" : true,
    "enabled" : false
},
{
    "id" : "82d3ec60-f2c0-4965-9825-cf88c59b4c3a"
    "name" : "manually-created-policy",
    "providerId" : "client-policy-provider",
    "description" : "This policy is created for the first test for introducing Client Policy.",
    "preSet" : false,
    "enabled" : true,
    "conditions" : [{
        "id" : "047eb7c3-ba5f-45bc-9e27-9ca0d3ae5146",
        "name" : "ClientRolesCondition",
        "providerId" : "clientroles-condition",        
        "config" : {"roles" : [ "sample-client-role" ]}
    }],
    "executors" : [{
        "id" : "47e818e3-f0d8-4a90-ace7-1a8b2362cec7",
        "name" : "SecureRequestObjectExecutor",
        "providerId" : "secure-reqobj-executor",        
        "config" : {}
    }]
}
```

- In the admin console, we need the ability to show also the executors and conditions of the "preSet" policies. Hence for the REST endpoint for read policies, it will likely need the ability to return all of the informations (including executors and conditions of the particular policy etc). However the REST endpoint for edit may just need the capability for enable/disable the preSet policy and nothing else should be updateable.

tnorimat:
@mposolda I agree with all points you've mentioned. 
There is one point I'd like to mention. As for adding "preSet" and "enabled" fields to `ComponentModel` and `ComponentRepresentation`, how about putting these information onto existing "config" fields? To do so, we need not to modify `ComponentModel` and `ComponentRepresentation`.

Some existing provider represented by `ComponentModel` and `ComponentRepresentation` have already had "enabled" field under "config" field. If adding "enabled" fields to `ComponentModel` and `ComponentRepresentation`, the representation of enabled/disabled information is doubled.

E.g. KeyProvider 
```
 {
                "name": "PS256",
                "providerId": "rsa-generated",
                "subComponents": {},
                "config": {
                    "keySize": [
                        "2048"
                    ],
                    "active": [
                        "true"
                    ],
                    "priority": [
                        "100"
                    ],
                    "enabled": [
                        "true"
                    ],
                    "algorithm": [
                        "PS256"
                    ]
                }
            },
            {
                "name": "ES256",
                "providerId": "ecdsa-generated",
                "subComponents": {},
                "config": {
                    "ecdsaEllipticCurveKey": [
                        "P-256"
                    ],
                    "active": [
                        "true"
                    ],
                    "priority": [
                        "100"
                    ],
                    "enabled": [
                        "true"
                    ]
                }
            }
```

tnorimat:
@mposolda As for the "enabled" claim for enabling/disabling the policy, it might be useful for the keycloak administrator to enable/disable the condition and executor. WDYT?

tnorimat:
@mposolda As for the "preSet" claim for indicating the policy being pre-set policy, it might be good to mark pre-set condition and executor with this "preSet" claim. 

The reason why is as follows.

When deleting the condition/executor individually, keycloak needs to check whether this condition/executor is under the pre-set policy or not. 

The condition and executors does not have the information on which policy includes it. Therefore, keycloak needs to look into policies sequentially to find which policy includes the condition/executor to be deleted. It cost a lot.

By introducing the marker indicating that the condition/executor is under the pre-set policy , we can avoid this costly operation.

mposolda:
@tnorimat We discuss with @stianst on how to move forward.

The preferred way for the client policies would be to have the JSON saved as the realm attribute and have the content of this JSON smaller and directly containing properties of the particular policy/condition/executor. So rather something as a blob. Something similar as described in the User profile design https://github.com/keycloak/keycloak-community/blob/master/design/user-profile.md#default-provider . 

Format can possibly look like this for our current policies implementations:

```
{
    "policies": [{
        "name": "default",
        "conditions": {
                "clientroles-condition": {
                    "roles", ["admin", "user"]
                },
                "client-ipaddr-condition": {
                    "ipaddr", ["10.20.30.40", "50.60.70.80"]
                }                
        }
        "executors": {
            "pkce-enforce-executor": {},
            "secure-session-enforce-executor": {}
        }        
    }]
}
```

So instead of using component model and the format of the JSON tight to the component model, we can use this one. In the admin console, we can have just the JSON editor for the first RHSSO 7.5 release.

WDYT?

stianst:
To add to what Marek said. We are looking towards improving REST APIs in the future. The component model, and magnitude of endpoints are just way to complex, and hard to manage. Rather, we are looking at APIs more akin what Kubernetes has. So for client policies there'd be something like:

/client-policies   -- List all client policies
/client-policies/{policy}  -- CRUD on a client policy

One thing I'm wondering in that regards is if we should make a split between "conditions" and "executors". I can imagine that'd we want to separate those two things. For example we could have a built-in hard-coded "FAPI profile", which essentially is just the list of executors, an admin can then use an exiting "profile" and set the conditions it will apply to. So perhaps it's more something like:

/client-policies/policies
/client-policies/policies/{policy}
/client-policies/profiles
/client-policies/profiles/{profile}

Where a "profile" would be a list of executors, while a policy would be a list of conditions and refer to one (or more?) profiles.

mposolda:
@stianst The dividing to "policies" and "profiles" makes sense to me.

Few more things to clarify:

- How exactly we save the policies and profiles? The possible options are:
1) Have single realm attribute, which will have JSON string of the list of all "policies" . Similarly have single attribute to have JSON of the list of all "profiles"
2) Have separate attributes - one attribute for each "policy" and one for each "profile"

With the approach (2), it will be harder to obtain list of all the policies. Unless, we have some convention like the attributes will need to have the prefix. So we would have something like "client.policies.fapi" and "client.policies.custom-policy" for the case when administrator want to have 2 policies called "fapi" and "custom-policy" . This looks a bit hacky, so my vote is to go with the approach (1) and just use the single attribute for handle the list of policies. And similarly for the profiles. WDYT?

- How to save the preSet (builtin) profiles? I assume, we will have builtin profile "fapi" . So since this is built-in thing, we don't need to save this "fapi" profile at all, right? So for example if administrator configures 2 additional profiles called "profile1" and "profile2", then all the available profiles in the realm will be "fapi", "profile1" and "profile2" . In other words, profile "fapi" will be always available and administrator won't be able to update "fapi" profile. If he wants to do some configuration updates, he will need to create separate profile with the changes. WDYT?
 wants to do some configuration updates in some executors, he will need to create separate profile with the changes.

stianst:
Not sure how we should save it, and we should align with how we do it for user profiles. So talk to @pedroigor and figure something out :)

Yes, so I'd like to get away from saving default stuff (including auth flows) in the db. These can just be hardcoded in the codebase. Perhaps this is some JSON files included in the dist? Maybe even it could make sense to have a directory on KC server where we can have client policies and user profiles on a global level?!? Just brainstorming on that last part ;)

mposolda:
@stianst Regarding the user profile, there is nothing yet available in Keycloak master? Or is it some work already @pedroigor ?

I can summary my understanding how to proceed saving etc. We can align both client policies and user profile to same/similar behaviour regarding storage etc:

- The final result is the union of "built-in" and "configured" . So for example when configured profiles "foo-profile" and "bar-profile", the used profiles would be "fapi", "foo-profile" and "bar-profile" (The "fapi" is built-in, others are configured)

- There will be 2 realm attributes: "client-policies.policies" to store the array of policies i JSON string. "client-policies.profiles" to store the array of profiles in JSON string

- Attributes will be empty by default. Which means that only built-in profiles ("fapi") will be used. There won't be policies by default

- I vote for have the built-in profiles either in the JSON as proposed by @stianst. They can be also hardcoed inside Java code, but JSON is likely better. IMO it will be nice if there is "automatic" migration of the built-in stuff. For example if we decide that we want to add new executor "new-executor" in Keycloak 16, then  people migrating from Keycloak 15 to Keycloak 16 will automatically see the new built-in "new-executor" inside the "fapi" profile. Will be nice if built-in JSON is hidden (bundled inside some JAR?), so that people don't have opportunity to break it during migration (EG. avoid the similar like "themes" directory which can people easily break during the migration).

WDYT?

pedroigor:
@mposolda JSON should be fine and that is what the contributors behind user profile are willing to do.

mposolda:
@tnorimat What do you think? Will it be possible to slightly change the design to align with the above?

The example of the "policies" can be maybe like:
```
{
    "policies": [{
        "name": "administrator-connect-from-local-network",
        "conditions": {
                "clientroles-condition": {
                    "roles", ["admin"]
                },
                "client-ipaddr-condition": {
                    "ipaddr", ["10.20.30.40", "192.168.0.0/16"]
                }                
        }
        "profiles": [ "foo", "bar" ]    
    }]
}
```
And profiles like:
```
{
    "profiles": [
    {
        "name": "foo",
        "executors": [
            "pkce-enforce-executor": {},
            "secure-session-enforce-executor": {}
        ]
    },
    {    "name": "foo",
         "executors": [
            "pkce-enforce-executor": {},
            "secure-session-enforce-executor": {}   
        ]
    }
  ]
}
```

tnorimat:
@mposolda Basically, I agree with the result of the discussion here. I will start revising the design document along with the result of the discussion.

I have a few comments as follows.

* [Modify merged condition/executor implementation provider/provider factory]
I think we need to modify merged condition/executor implementation provides/provider factories because they use `ComponentModel` to save their configuration. These configurations can be expressed as JSON representation stored in the realm's attribute. Therefore, merged condition/executor implementation provides/provider factories manage their configuration via this realm's attribute. Is that right?

* [Where to save JSON representation of built-in profiles]
As you suggested, I think we save JSON representation of built-in profiles on `keycloak-services`'s JAR files to prevent the keycloak administrators from modifying this JSON representation. WDYT?

* [Enable/Disable policy] As you proposed, I think we need to enable/disable policies. WDYT?

* [Enable/Disable profile] I'm not sure whether we need to enable/disable profiles. Instead of disabling the profile, we can modify the policy not to refer to this profile. Moreover, If we are able to enable/disable built-in profiles, we need to persist the information on whether the built-in policy is enabled or disabled, which cause the issue on where it is persisted.  Therefore, IMO, we need not to enable/disable profiles. WDYT?

* [Enable/Disable conditions and executors] I'm also not sure whether we might be able to enable/disable conditions and executors. Could you give me some idea on it?

* [Impact on the on-going Concept Design for Client Policies] IMO, the result of the discussion on this PR does not affect so much on the existing Concept Design https://marvelapp.com/prototype/6e70eh2/screen/74918976 . We need to add the UI for the editor of JSON representation of the whole Client Policies. Is it right?

mposolda:
@tnorimat These are good questions. We had some more discussion with @stianst and @pedroigor in this topic. I will try to summary what we discuss.

In basics, the idea is Keycloak is secured by default. Hence for example there won't be allowed to configure redirect URL like "*" or create clients with "Full Scope Allowed" enabled etc. Similarly for example PKCE should be enabled/enforced by default for the public clients etc.

For this, we may have the profile like "oauth-best-practices" , which will contain some basic set of executors to enforce this. We may have more built-in profiles defined (For example the "fapi-read" profile and "fapi-rw" profile), which will have some different set of executors defined on them. However those "Fapi-*" profiles won't be used by default - administrator will need to bind them to some defined policy with some set of conditions to be applied to.

So as you pointed, we may have some policy like "default", which will be enabled and used by default. Policy may just not need have any conditions, so it will be always applied (or alternatively it can have "always-use" condition on it). So the configuration can be like this:
```
policy {
    "conditions": [ "always-use" ],
    "enabled": true,
    "profiles": [ "oauth-best-practices" ]
}
```

Administrator will have a chance to disable this and in that case, it will need to be changed as the attribute on the realm (By default, the built-in policies and profiles don't need to be saved anywhere.) More inline:

> 
> I have a few comments as follows.
> 
>     * [Modify merged condition/executor implementation provider/provider factory]
>       I think we need to modify merged condition/executor implementation provides/provider factories because they use `ComponentModel` to save their configuration. These configurations can be expressed as JSON representation stored in the realm's attribute. Therefore, merged condition/executor implementation provides/provider factories manage their configuration via this realm's attribute. Is that right?
> 

Yes, I think we may need to modify existing items to match with this, so that they don't use ComponentModel.

You don't need to care about DB migration. For example if your changes are merged to Keycloak master after Keycloak 12 release, it means that Keycloak 12 would be released with ComponentModel used. So people migrating from Keycloak 12 to Keycloak 13 will need to re-create their client policies (no migration support provided). That is fine because Client policies are still "preview" and hence we don't care about backwards compatibility.

>     * [Where to save JSON representation of built-in profiles]
>       As you suggested, I think we save JSON representation of built-in profiles on `keycloak-services`'s JAR files to prevent the keycloak administrators from modifying this JSON representation. WDYT?

+1 , I think that saving the JSON directly inside `keycloak-services` JAR will be good. Maybe something like `keycloak-default.json` ?

> 
>     * [Enable/Disable policy] As you proposed, I think we need to enable/disable policies. WDYT?

+1 as mentioned above

> 
>     * [Enable/Disable profile] I'm not sure whether we need to enable/disable profiles. Instead of disabling the profile, we can modify the policy not to refer to this profile. Moreover, If we are able to enable/disable built-in profiles, we need to persist the information on whether the built-in policy is enabled or disabled, which cause the issue on where it is persisted.  Therefore, IMO, we need not to enable/disable profiles. WDYT?

Yes, I think that your proposal is nice and we don't need the ability to enable/disable profiles.

> 
>     * [Enable/Disable conditions and executors] I'm also not sure whether we might be able to enable/disable conditions and executors. Could you give me some idea on it?

I think we don't need this as we can handle this by just enable/disable policies. For example, if someone wants to enable/disable executor from the existing profile (like for example from "fapi" profile), he may just need to "copy" the existing "fapi" profile and then just remove the target executor from it. A bit similar manner like authenticationFlows when you need to "copy" the flow if you want to add/remove authenticationExecution from it and then you need to "bind" the copy of that flow.

> 
>     * [Impact on the on-going Concept Design for Client Policies] IMO, the result of the discussion on this PR does not affect so much on the existing Concept Design https://marvelapp.com/prototype/6e70eh2/screen/74918976 . We need to add the UI for the editor of JSON representation of the whole Client Policies. Is it right?

Yes, I think that in the first version, we may need just the ability to CRUD the JSON file directly in the admin console. And this is something, which can be probably relatively easily added to the old admin console as well for now? Just add the new tab "Client policies" next to the tab "Client Registration Policies" and just add the JSON CRUD support in there for policies and CRUD support for profiles - the built-in profiles are not editable. The built-in policies can be just enabled/disabled but not re-configured.

Regarding more fine-grained UXD design in the new admin console, we may need to come back to the UXD team once we have more clear this design (and maybe the initial implementation of JSON in the old admin console? I am not sure...) WDYT?

tnorimat:
@mposolda Thank you for your comments. I agree with you generally.

I have some following comments.

* [term : preset or built-in]
 We might use the term "preset" and "built-in" interchangeably to represent "something that has already been included and not modified". Could we choose the term "preset" or "built-in" to represent this concept? I would like to give a vote o "built-in". On this comments, I use the term "built-in".

* [policy's human readable description]
 Could we include the "description" field to the policy? It can be useful for the keycloak administrator to recognize what this policy is (especially, on Admin Console UI).

* [profile's human readable description]
 Could we include the "description" field to the profile? It can be useful for the keycloak administrator to recognize what this profile is (especially, on Admin Console UI).

* [how to persist built-in policy]
 As you mentioned, we could enable/disable the built-in policy. 
 Policies and profiles can be stored on the following locations.
   - Hard coded in binary
    Policies and profiles that are not able to be modified in their nature should be stored here.
   - Realm attribute
   Policies and profiles that are able to be modified in their nature should be stored in the realm attribute.

  Considering these points, ordinal policies and profiles should be stored in the realm attribute while built-in profiles should be stored in the hard coded in binary.
 How about a built-in policy? It include both parts that can be modified and not. "enabled" filed of the built-in policy can be modified while others can not be modified.
 My proposal for storing the built-in policy is as follows.
   * Basically, the part that can be modified is stored and managed in the realm attribute while the other in the realm attribute.
   * The policy have "builtIn" field to show whether this policy is built-in policy or not.
   * The whole built-in policy is stored and managed in the hard coded in binary. This policy's "enabled" field is fixed as "true".
   * Only "name" and "enabled" fields are stored and managed in the realm attribute. 
   * "enabled" field of the build-in policy stored in the realm attribute overrides the one stored in the hard coded in binary.

* [UX Design]
 As you mentioned, we need to discuss the design with UXD after clarifying this specification.

mposolda:
> @mposolda Thank you for your comments. I agree with you generally.
> 
> I have some following comments.
> 
>     * [term : preset or built-in]
>       We might use the term "preset" and "built-in" interchangeably to represent "something that has already been included and not modified". Could we choose the term "preset" or "built-in" to represent this concept? I would like to give a vote o "built-in". On this comments, I use the term "built-in".

+1 for built-in

> 
>     * [policy's human readable description]
>       Could we include the "description" field to the policy? It can be useful for the keycloak administrator to recognize what this policy is (especially, on Admin Console UI).

+1

> 
>     * [profile's human readable description]
>       Could we include the "description" field to the profile? It can be useful for the keycloak administrator to recognize what this profile is (especially, on Admin Console UI).

+1

> 
>     * [how to persist built-in policy]
>       As you mentioned, we could enable/disable the built-in policy.
>       Policies and profiles can be stored on the following locations.
>       
>       * Hard coded in binary
>         Policies and profiles that are not able to be modified in their nature should be stored here.
>       * Realm attribute
>         Policies and profiles that are able to be modified in their nature should be stored in the realm attribute.
>       
>       Considering these points, ordinal policies and profiles should be stored in the realm attribute while built-in profiles should be stored in the hard coded in binary.
>       How about a built-in policy? It include both parts that can be modified and not. "enabled" filed of the built-in policy can be modified while others can not be modified.
>       My proposal for storing the built-in policy is as follows.
>       
>       * Basically, the part that can be modified is stored and managed in the realm attribute while the other in the realm attribute.
>       * The policy have "builtIn" field to show whether this policy is built-in policy or not.
>       * The whole built-in policy is stored and managed in the hard coded in binary. This policy's "enabled" field is fixed as "true".
>       * Only "name" and "enabled" fields are stored and managed in the realm attribute.
>       * "enabled" field of the build-in policy stored in the realm attribute overrides the one stored in the hard coded in binary.

Yes, this proposal makes a lot of sense to me. I had also some thoughts that it can be likely done this way. So +1 for doing it like this.

>     * [UX Design]
>       As you mentioned, we need to discuss the design with UXD after clarifying this specification.

+1

tnorimat:
@mposolda Thank you for your comments. It seems that all discussion points are clarified so that I will revise the design document by incorporating these points.

mposolda:
@tnorimat Cool, Thanks!

tnorimat:
@mposolda I've updated the document. Could you check whether the updated version of this document incorporates all what we have discussed? If this PR is merged, I'll start implementing new Admin REST APIs for Client Policies by following this document.

design/client-policies.md
```diff
@@ -207,16 +224,472 @@ Conditions and executors of a policy should be evaluated on endpoints where a cl
...
+To add/read/modify/delete conditions/executors/policies/profiles, following Admin REST APIs are supported.
+
+The style of API description conforms to [keycloak's API Documentation](https://www.keycloak.org/docs-api/11.0/rest-api/index.html).
+
+### Design Concept
+
+Admin REST API for Client Policies are designed to satisfy the following requirements.
+
+- keycloak admin can view and update the client policy on the editable textarea UI (editing JSON representation directly).
+- keycloak admin can control policies, conditions and executors on the ordinal UI (e.g. like the existing Client Registration Policy).
```
mposolda, line 256: Just a note, that there is a question how to implement these points because the new admin console is in the design right now, but it is not yet ready... So if you want to start implementation of these points, you may not yet be able to add support in the admin console UI of new admin console as it is not ready yet... 

If you implement some support to the old admin console, I suggest to add here only the editing by JSON to be able to test client policies with the old admin console.

mposolda (approved):
@tnorimat Thanks a lot! I am approving and merging your PR.

I've added just one inline comment regarding the old admin console: Basically knowing that old admin console will be removed soon, I suggest you to not spend much time on tweaking the old admin console. If needed, only add the parts related to JSON edit, but maybe even that is not needed at all... As long as we have REST endpoints, we are good to go.

design/client-policies.md
tnorimat: @mposolda Thank you for your comment.

AFAIK, the situation for the New Admin Console are as follows.

- The New Admin Console has not yet appeared on the latest keycloak.
- The UI design for Client Policies on the New Admin Console has not yet been fixed. It has been pending because the design for the New Admin REST API for Client Policies had not yet been approved and merged.

Considering these points, I would like to work on the support of New Admin REST API as follows.

1. Resume discussion on the UI design for Client Policies on the New Admin Console (https://marvelapp.com/prototype/6e70eh2/screen/74918976) and fix it.
2. In parallel, Implement the New Admin REST API for Client Policies by following this merged design document.
3. After completing 1 and 2, if the New Admin Console appears on the keycloak's master branch, implement the UI for Client Policies on the New Admin Console. If not, implement the UI for Client Policies on the existing Admin Console (only editable JSON).

mposolda: @tnorimat Your proposal sounds great to me. Thanks!

ssilvert: @tnorimat FYI.  The new admin console is being developed in its own repo: https://github.com/keycloak/keycloak-admin-ui

I agree with @mposolda that your proposal is good.  If you want this functionality soon then it will need to go into old admin console.

tnorimat: @ssilvert Thank you for your comment. I would like to check https://github.com/keycloak/keycloak-admin-ui .

PR #190, opened 2020-11-12
KEYCLOAK-11019 Offline Session Lazy Loading

y-tabata:
This PR is for [KEYCLOAK-11019 Offline sessions preloading still slow in some cases](https://issues.redhat.com/browse/KEYCLOAK-11019).
In a large number of Offline Sessions environments, we can shorten the startup time by tuning some parameters or adding indexes, but there is a limit we can shorten.
In this PR, we'd like to consider resolving the root cause, that is, consider changing the offline session loading method from preloading all offline sessions in the startup, and consider something like "Offline Session Lazy Loading".

### DIFF
--- a/design/offline-session-lazy-loading.md
+++ b/design/offline-session-lazy-loading.md
@@ -1,0 +1,44 @@
+# Offline Session Lazy Loading
+
+* **Status**: Draft #1
+* **JIRA**: [KEYCLOAK-11019](https://issues.redhat.com/browse/KEYCLOAK-11019)
+
+## Motivation
+
+To make it fast to startup Keycloak with a large number of Offline Sessions.
+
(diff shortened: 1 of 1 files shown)


pedroigor:
@y-tabata I also think the moment requires being a bit conservative and hold drastic changes for later once we discuss the overall design on how we store and manage sessions.

So, back to your proposal:

* I agree to have a switch for enablig/disabling pre-fetch. I don't like it very much but it should help for now to solve the problem while still supporting the current behavior. I think we should disable by default, document this change, wait for feedback, work on improvements, and then finally remove pre-fetch from startup.

* Maybe we could start a background process during startup to fetch offline sessions from the database. Maybe this should help to avoid changing code too much with a possible impact on user experience where sessions would be eventually restored as they are loaded from the database

* I remember @mposolda mentioning some impacts on the admin console where you need session counts, etc. I'm OK to sacrifice that if the user wants too. For me, the benefits justify the risks and users can always choose what works better for them by using the switch.

thomasdarimont:
Just wanted to link this to the [PR 7722](https://github.com/keycloak/keycloak/pull/7722) which is currently WIP.

hmlnarik:
Closing this PR as outdated. The PR would have to contain description of the approach to be mergeable. The keycloak/keycloak#8033 (based on keycloak/keycloak#7722) has implemented lazy offline session loading and for this PR to be mergeable, it would need to be aligned with the code PR. If desired, feel free to reopen this PR with the details describing the implementation.

PR #213, opened 2021-01-13
Token Exchange SPI draft

dteleguin:
Token Exchange SPI

### DIFF
--- a/design/token-exchange-spi.md
+++ b/design/token-exchange-spi.md
@@ -1,0 +1,86 @@
+# Token Exchange SPI
+
+## Abstract
+This document outlines the design of the proposed Token Exchange SPI for Keycloak (TODO: JIRA link).
+## Rationale
+Keycloak does implement the OAuth 2.0 Token Exchange ([RFC 8693](https://tools.ietf.org/html/rfc8693)), but does that in a peculiar way (Securing Applications and Services Guide, [7. Token Exchange](https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange)):
+> Token exchange in Keycloak is a very loose implementation of the OAuth Token Exchange specification at the IETF. We have extended it a little, ignored some of it, and loosely interpreted other parts of the specification.
+
+Current implementation has the following limitations and shortcomings:
+* The following (optional) parameters are not supported:
+	* `resource`
+	* `scope`
+	* `actor_token` and `actor_token_type`
+* Supporting internal exchange to/from non-OAuth clients requires core Keycloak modification. Currently, only exchange to SAML is supported;
+* Exchanging to/from external IdPs is supported for OAuth IdPs only and requires brokering to be set up;
+* Token exchange is built into `TokenEndpoint` which is already ~1400 LoC.
+
+Considering the above, we are suggesting the introduction of the Token Exchange SPI.
+
+## Use Cases and Benefits
+* Complement Protocol SPI: allow to support exchanging to/from non-OAuth clients without modifying Keycloak
+* Complement Identity Provider SPI: allow to support exchanging to/from non-OAuth IdPs without modifying Keycloak
+* Non-brokered trust: allow to recognize, validate and exchange tokens issued by 3rd party entities in the cases when the trust cannot be expressed in terms of brokering
+
+The SPI will also allow to clean up `TokenEndpoint` and to improve its maintainability.
+
+## Design
+* `TokenExchangeSPI`: SPI class
+* `TokenExchangeProviderFactory`: factory interface
+* `TokenExchangeProvider`: main (provider) interface; should define the `exchange()` method:
+```
+AccessTokenResponse exchange(TokenExchangeContext);
+```
+* `TokenExchangeContext`:
+	* should supply the data needed to perform exchange (parsed request parameters);
+	* should provide access to `KeycloakSession`;
+	* may also contain auxiliary helper objects like `TokenManager`.
+
+### Configuration
+
+#### Provider level
+Token exchange providers could be configured using the standard `org.keycloak.Config` API. The configuration will be global, and will be stored in the filesystem (or whichever config store is selected).
+
+#### Instance level
+In addition to that, we can provide instance-level configuration, similar to authenticators. The configuration will realm-local, and will be stored in the database. The relevant UI parts will need to be added to the Admin Console.
+
+### Provider selection
+Based on the request data, one and only one token exchange provider should be selected. There are several options for that:
+
+#### Provider-side selection
+The selection process could be delegated to the providers themselves. For that, `TokenExchangeProvider` could define the additional `supports()` method:
+```
+boolean supports(TokenExchangeContext);
+```
+This method alone won't be sufficient, since any two providers can return `true` simultaneously. To resolve the possible conflict, we'll need to introduce deterministic ordering of the providers (e.g. via the `TokenExchangeProviderFactory#getPriority` method).
+
+This option assumes that the provider selection should be the responsibility of the provider developer (and not end user).
+
+#### Policy-based selection
+As a different option, the selection process could be decoupled from the provider code completely. For that, we should introduce policies (predicates) that would operate on the request data and return a decision; each policy will be associated with a single provider or a processor (configured provider instance).
+
+This option assumes that the provider selection should be configurable by end users (and thus would require a UI).
+
+## Implementation plan
+### Phase 0
+(Backbase internal PoC)
+### Phase 1
+* Add Token Exchange SPI interfaces and classes
+* Refactor `TokenEndpoint` to use the new SPI
+* Move Keycloak's token exchange implementation to `DefaultTokenExchangeProvider`
+* Implement provider-based selection
+### Phase 2
+* Convert Keycloak's built-in token exchange methods to the new SPI
+* Use Grant Type SPI
+### Phase 3
+* Implement policy-based provider selection
+* Implement token exchange processors (configured provider instances)
+## Relation to other SPIs
+### Grant Type SPI
+The RFC says:
+> The client makes a token exchange request to the token endpoint with
+   an extension grant type using the HTTP "POST" method.
+
+Thus, token exchange is presumed to be implemented as extension grant (see RFC 6749, The OAuth 2.0 Authorization Framework, [4.5. Extension Grants](https://tools.ietf.org/html/rfc6749#section-4.5)). Once Keycloak has Grant Type SPI and proper extension grants support, the whole token exchange feature should be ported to the new SPI.
+## Relation to other tasks/epics
+[KEYCLOAK-8276: Review Token Exchange](https://issues.redhat.com/browse/KEYCLOAK-8276)
--- a/design/token-exchange-spi.md
+++ b/design/token-exchange-spi.md
@@ -1,12 +1,18 @@
 # Token Exchange SPI
 
+* **Status**: Notes
+* **JIRA**: _TBD_
+
 ## Abstract
-This document outlines the design of the proposed Token Exchange SPI for Keycloak (TODO: JIRA link).
-## Rationale
+This document outlines the design of the proposed Token Exchange SPI for Keycloak.
+
+## Motivation
 Keycloak does implement the OAuth 2.0 Token Exchange ([RFC 8693](https://tools.ietf.org/html/rfc8693)), but does that in a peculiar way (Securing Applications and Services Guide, [7. Token Exchange](https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange)):
+
 > Token exchange in Keycloak is a very loose implementation of the OAuth Token Exchange specification at the IETF. We have extended it a little, ignored some of it, and loosely interpreted other parts of the specification.
 
 Current implementation has the following limitations and shortcomings:
+
 * The following (optional) parameters are not supported:
 	* `resource`
 	* `scope`
@@ -24,7 +30,7 @@ Considering the above, we are suggesting the introduction of the Token Exchange 
 
 The SPI will also allow to clean up `TokenEndpoint` and to improve its maintainability.
 
-## Design
+## Implementation details
 * `TokenExchangeSPI`: SPI class
 * `TokenExchangeProviderFactory`: factory interface
 * `TokenExchangeProvider`: main (provider) interface; should define the `exchange()` method:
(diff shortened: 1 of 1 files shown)


design/token-exchange-spi.md
```diff
@@ -0,0 +1,86 @@
...
+	* `resource`
+	* `scope`
+	* `actor_token` and `actor_token_type`
+* Supporting internal exchange to/from non-OAuth clients requires core Keycloak modification. Currently, only exchange to SAML is supported;
+* Exchanging to/from external IdPs is supported for OAuth IdPs only and requires brokering to be set up;
+* Token exchange is built into `TokenEndpoint` which is already ~1400 LoC.
+
+Considering the above, we are suggesting the introduction of the Token Exchange SPI.
+
+## Use Cases and Benefits
```
pedroigor, line 20: I think we should group use cases accordingly to the different token exchange scenarios they are associated with:

* **Impersonation**: Not sure if covered in this section but one of the use cases we support. This one is probably what resource servers are mainly interested in in order to call other services assuming the role of the original client to which the token was issued for. The current implementation needs to be checked whether it is respecting the semantics for this scenario when issuing tokens. For instance, the `client_id` claim.
* **Delegation**: This is the scenario that I think is the most problematic in terms of compliance with specs. Even in docs we talk about delegation in terms of impersonation, where, for RFC 8693, delegation is related to the subject delegating to an actor to act on his behalf. At this regard, our current implementation lacks a lot of the semantics defined by the spec.
* **Cross-domain**: This is the scenario where tokens are exchanged to/from different domains. Currently, we have this trust scenario solely based on brokering, where you can exchange an internal to external, and external to internal tokens.

Your use cases seem to be related to the **Trust** scenario where:

* Complement Protocol SPI: non-OAuth/OIDC clients are by nature outside the boundaries of a realm.
* Complement Identity Provider SPI: non-Auth Idps are outside the boundaries of a realm (similar to brokers)
* Non-brokered trust: Probably the use case closer to the **Trust** scenario I'm mentioned. However, I'm not sure about the requirements around recognizing and validating tokens but exchanging.

For me, looks like we need to consider all those scenarios prior to discussing a re-design and new SPIs. All the use cases you defined seem to be related to trust. And I think the other scenarios might (or not) impact any decision we make about the design.

design/token-exchange-spi.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Abstract
+This document outlines the design of the proposed Token Exchange SPI for Keycloak (TODO: JIRA link).
+## Rationale
+Keycloak does implement the OAuth 2.0 Token Exchange ([RFC 8693](https://tools.ietf.org/html/rfc8693)), but does that in a peculiar way (Securing Applications and Services Guide, [7. Token Exchange](https://www.keycloak.org/docs/latest/securing_apps/index.html#_token-exchange)):
+> Token exchange in Keycloak is a very loose implementation of the OAuth Token Exchange specification at the IETF. We have extended it a little, ignored some of it, and loosely interpreted other parts of the specification.
+
+Current implementation has the following limitations and shortcomings:
+* The following (optional) parameters are not supported:
+	* `resource`
```
pedroigor, line 11: What are your plans about the `resource` parameter? Isn't the audience enough and potentially easier to handle because it is a "logical" name of the targeted service?

I'm wondering if you have some specific requirement to choose the proper audience/target service based on a specific URL to where the token should be created for.

design/token-exchange-spi.md
```diff
@@ -0,0 +1,86 @@
...
+	* `scope`
+	* `actor_token` and `actor_token_type`
```
pedroigor, line 13: These are related to the delegation scenarios covered by the specs that in our docs/implementation we refer to as impersonation.

From the list of use cases herein provided, is delegation among them?

pedroigor (commented):
Hey @dteleguin, some initial thoughts. Thanks for the great document.

In general, I think we should make clear first the use cases we want to support and build a plan based on it. For me looks like you are more interested to support trust scenarios where Keycloak can easily be configured/extended to exchanged tokens from different formats and from different security domains.

Please, let me know what you think.

design/token-exchange-spi.md
jonm-bb: The plan is to implement all of the parameters supported by the spec, but by delegating the values to the registered Providers so that they can inspect them and elect to handle this exchange. From this perspective, the `resource` parameter is just one of the options for a Provider to use to determine if it should be the one to handle this exchange request.

design/token-exchange-spi.md
jonm-bb: The biggest use case pushing us towards making this change is delegation, where the customer of a financial institution can delegate access to their accounts and data to a customer services representative from the FI. In this case, the customer is the subject of the requested token, and the CSR is the actor that will be using the token. In such scenarios, we have found that the customers and employees are sometimes using different OPs/IdPs and are therefore expected to use access tokens in different formats with different issuers. For an employee (who has a Ping token) to exchange that token for a customer (Keycloak) token without going to Ping to perform re-authentication, it should be possible to exchange the Ping token for a Keycloak token that has the appropriate claims embedded (e.g. the `act` claim identifying the employee)

design/token-exchange-spi.md
jonm-bb: I'd agree that our first use case is about cross-domain trust, but with elements of delegation thrown in for good measure. 
However, the idea behind the SPI is to go beyond this and make token exchange as flexible as possible, so that tokens can be exchanged for a variety of purposes. For example a token issued by a (non-standard) legacy system could be exchanged for a Keycloak token that allows access to a more modern API or system; or an access token can be exchanged at the system edge for an internal format understood by the protected API - so long as there is a Token Exchange SPI Provider loaded that understands the intent based on the provided parameters (audience, token type etc), then a wide variety of use cases should be possible.

pedroigor: Looks like what we need then is not really a Token Exchange SPI but some SPI specific for token exchange that allows supporting different token types.

Based on your comment, I think we could:

* Implement a `TokenProvider`, where the implementation is targeted for parsing a specific token type (e.g.: `subject_token_type` == `urn:mytoken`)
* Parse, validate and extract information from the `subject_token`. Here also lies the place where a provider can be used for cross-domain trust.
* Create a generic representation for the incoming token
* Elect a `TokenProvider` based on the `requested_token_type` (or maybe another interface specific for this flow)
* Invoke methods on the elected implementation to transform/exchange the representation into the requested token type

Considering the current implementation, the code we have today to generate the token response for the token types we support would be within a specific token provider implementation (and not hardcoded into the token endpoint). The same goes for the code parsing the `subject_token`.

Note, however, that token exchange is flexible in regards to the response.

```
access_token
      REQUIRED.  The security token issued by the authorization server
      in response to the token exchange request.  The "access_token"
      parameter from Section 5.1 of [RFC6749] is used here to carry the
      requested token, which allows this token exchange protocol to use
      the existing OAuth 2.0 request and response constructs defined for
      the token endpoint.  The identifier "access_token" is used for
      historical reasons and the issued token need not be an OAuth
      access token.
```

So we can include whatever token we want into the `access_token` response claim.

I think something like that makes more sense than a Token Exchange SPI.

pedroigor:
```
interface TokenExchangeProviderFactory {
    
   /**
   * Returns the token types that can be handled by this provider.
   */
    String[] getSupportedTokenTypes();

   /**
   * Returns the token types that can be issued by this provider.
   */
    String[] getSupportedRequestedTokenTypes();

    /**
     * Returns the actor token types supported by this provider for delegation scenarios.
     */
    String[] getSupportedActorTokenTypes();
}

interface TokenExchangeProvider {

     /**
      * Validates the token request and returns a {@link TokenExchangeContext} holding the information
      * from the token that should be exchanged.
      */
     TokenExchangeContext validate(TokenRequest) throws TokenExchangeException;

     /**
      * Perform the token exchange based on the information from the {@link TokenExchangeContext}.
      */
     TokenExchangeResponse exchange(TokenExchangeContext);
}

/**
* Represents the information associated with a token exchange request
*/
interface TokenExchangeRequest {
    // access to token exchange request params
}

/**
* The context represents a valid token request, providing additional information that can be used by other providers
* to exchange a subject token to a requested token type.
*
* It should be possible to have multiple sub-types of this interface representing a specific token exchange context. For instance, an OAuth2TokenExchangeContext representing an OAuth2 subject token that can be exchanged for another (or the same) token type.
*/
interface TokenExchangeContext {
   
    // information about the audience
    // information about the resource
    // information about the scope
    // information about the subject_token, which should include the subject, and the client asking for the exchange, perhaps the original issuer, etc.
    // information about the actor trying to act on behalf of the subject, if in a delegation scenario
}

interface TokenExchangeResponse {
   void setIssuedTokenType(String)
   void setIssuedToken(String);
   void setTokenType(String)
   void setExpiration(int);
   void setRefreshToken(String);
   void setScope(String);
}

This API should allow us to support token types depending on the use case. For instance, we would provide OOTB an `OAuth2TokenProvider` that is capable of validating and exchanging Keycloak tokens. If you want to support exchanging an internal token with an OAuth2 token, then you just need to implement the `validate` method in another provider implementation, leveraging the logic for generating/exchanging OAuth2 tokens to the OOTB provider.

The same goes if you want to exchange an OAuth2 token with an internal token. In this case, you just need a provider that implements the `exchange` method to build your internal token, leveraging the validation and representation from the default OAuth2 token provider.

Hope it makes sense. Like I said, quite similar to the API herein proposed. But with the possibility to re-use providers depending on the use case.

design/token-exchange-spi.md
pedroigor: I see. I would say that using the `resource` like that looks too fine-grained? I would just use both `audience` and `resource` parameters to influence how tokens should be issued, and not as a filter for choosing a provider?

design/token-exchange-spi.md
pedroigor: @jonm-bb This financial use-case you mentioned can be also solved using CIBA?

jonm-bb: CIBA would work from the target web application's perspective, which can contact Keycloak to trigger 'authentication', but as the employee will likely need to provide consent from one of the FI's web applications, it gets a bit awkward to contact for Keycloak to contact the web app for it to contact the user for them to 'authenticate' (by selecting which user they want to impersonate). For me, this use case is the wrong way round for CIBA - I would expect CIBA to be used when the person attempting to authenticate visits the app they are looking to authenticate into, which then contacts them over an appropriate channel. On top of that, the access token issued still needs to embed the `act` claim to identify the real end user. 
In this use case, the person being contacted (the CSR) is already authenticated elsewhere and needs to provide details of which customer they are being delegated access from; this is most seamless if they can simply exchange their access token for a limited one that will work with the target web app.
Other use cases where this token exchange mechanism would be advantageous is transparent token exchange at the system edge, where a token issued by the IdP is swapped for an internal format understood by the API being accessed, and CIBA is less useful there also.

design/token-exchange-spi.md
jonm-bb: I guess the purpose of the matching algorithm is to provide a simple way for one of the Providers to decide to be the one that handles the exchange. They are free to use the provided parameters in any way they deem appropriate - if they want to register an interest in a URL or a logical name, and are able to work with such, then I'd say it's fair game and there's no reason to limit the request parameters that a Provider can make use of.

design/token-exchange-spi.md
pedroigor: I see. I only mentioned CIBA because one of the use cases defined in the specs is related to an agent obtaining customer authorization to use his data for authentication/confirmation purposes. But yeah, not really delegation.

Thanks, for detailing the use case.

pedroigor:
@dteleguin @jonm-bb From my side, the goals are clear and I think we should just agree on the SPI to move forward.

dteleguin:
@pedroigor we have created and tested our internal PoC, and I'd say we're pretty much satisfied with it. The only concern is about the provider selection mechanism; I'm afraid that simple `{subject,requested,actor}_token_type` matching won't be sufficient here.

In our scenario, the whole subject token subclass had to be handled by a single provider, with the subclass being represented by a prefix (`subject_token_type=urn:backbase:*`).

If we also intend to modularize Keycloak's current token exchange (which I believe we do), we'll need to honor at least `requested_subject` (to distinguish impersonation from "normal" exchange) and `audience` (to distinguish internal and external exchange).

With the above, we have implemented the "Provider-side selection" scenario, which seems to be working perfectly. I think it should also work for Milestone 1.

pedroigor:
@dteleguin Sorry for the delay. Could you share some code?

dteleguin:
@pedroigor sure, please check out:
https://github.com/carrettiPro/keycloak/tree/token-exchange-spi

pedroigor:
@dteleguin Thanks. My only concern is how you would implement, for instance, support for exchanging an internal token with a OAuth2 token.

I guess the developer would need to deal with some code internals to implement that, right? The reason being that he would need to deal with all tricks we have for building these tokens (potentially creating sessions, etc).

dteleguin:
@pedroigor 

> I guess the developer would need to deal with some code internals to implement that, right?

That's correct; generating an access token would require quite a lot of boilerplate, starting with RootAuthenticationSession and ending with TokenResponse. I also remember people asking on the ML how to generate tokens from inside Keycloak (e.g. for an outgoing call from a KC provider to a service that would expect a KC-produced token). So I think it would be nice to have some helper methods in place for that.

> Perhaps you could provide an example of how people are going to use your SPI?

Sure, will do that shortly.

pedroigor:
> That's correct; generating an access token would require quite a lot of boilerplate, starting with RootAuthenticationSession and ending with TokenResponse. I also remember people asking on the ML how to generate tokens from inside Keycloak (e.g. for an outgoing call from a KC provider to a service that would expect a KC-produced token). So I think it would be nice to have some helper methods in place for that.

Yes, it would. And it would be nice if we could just reuse providers for specific token formats when doing exchanges.

That is what I tried to propose before, but I know it is not yet the best design for the SPI.

The overall idea would be that people trying to exchange an internal token with a Keycloak one (or even stored by brokers) would only require implementing the code to handle the internal token.

I think that would be really nice to see and more aligned with the purposes of the SPI.

dteleguin:
@pedroigor , I've published an [example](https://github.com/dteleguin/token-exchange-spi-examples) demonstrating simple cross-realm impersonation (without brokering) using Token Exchange SPI.

dteleguin:
As for your proposal, it is suggested that we (as the authors of token exchange handler) would implement token ingestion (parsing), then populate some minimal context (realm, client, user, audience, scope) and let some reusable token generator do the rest - is my understanding correct?

Moreover, I believe a full-fledged STS should have under the hood a configurable pipeline consisting of pluggable token parsers / generators / processors (transformers). Thus, the majority of token exchange use cases (including Keycloak's built-in ones) could be implemented via composing such building blocks with almost no coding.

At the same time, I think we should start with something basic and let the feature evolve in an iterative manner.

pedroigor:
@dteleguin That is exactly what I'm trying to propose. And I think that would be the long-term solution. Otherwise, it would be an SPI that could only be implemented by people with high knowledge about KC internals and still subject to errors (and unnecessary support tickets due to misuse of our not-so-good APIs).

I agree though that it would require a bit more discussion. And I also think your proposal helps to at least allow people to customize exchanges. Even though not the best solution.

I'm OK about considering the refactoring you did for now. But I would like to see @stianst opinion on this matter too.

pedroigor:
At the same time, our plans around Quarkus should help a lot with simplifications for our SPIs too.

PR #238, opened 2021-02-15
Keycloak-X Logging Design Document

thomasdarimont:


pedroigor (commented):
@thomasdarimont Thanks for opening this one. That is an important area in the context of troubleshooting.

In general, the proposal herein makes a lot of sense including the possibility to use different formats, starting with JSON.

I would like to propose some changes to the document so that we start linking things together.

I just sent a PR to the configuration design document with some updated to throubleshoting. Please take a look at https://github.com/keycloak/keycloak-community/pull/240.

My proposal is that we include in your proposal:

* Additional properties for logging and their purpose
* Additional CLI commands for logging (if necessary)
* Impact on the server distribution
* Examples of how users should configure logging using the CLI

And perhaps answering these questions when doing above:

* How to define a logging level?
* How to format log entries?
* How to change the log format?
* Shall we provide a `log` directory in the distribution?
* What is the default provider and settings for dev and production profiles?

By doing this, we should be able to agree on the top-level requirements, usage, and usability.

I also agree with you that we should start small, with a focus on what Quarkus provides, and then discuss integration with other providers/services

Please, let me know what do you think.

PR #254, opened 2021-03-24
DPoP design draft

dteleguin:


design/dpop.md
```diff
@@ -0,0 +1,259 @@
...
+
+It is suggested that DPoP should be set up on a per-client basis, with the three options available that should be interpreted as follows:
+
+|**Value** | **Keycloak Server** | **Keycloak Adaptors** |
+| --- | --- | --- |
+| Required | OAuth endpoints (where applicable) should require a DPoP header and a DPoP-bound token; should return an error otherwise. | Protected resources should require a DPoP header and a DPoP-bound token, or should return an error otherwise. |
+| Optional | _Token Endpoint_: if DPoP header is present, the returned token should be DPoP-bound and should contain a `cnf` claim with a JWK thumbprint; should be a regular token otherwise  <br/><br/>_UserInfo endpoint_: if the access token is DPoP-bound, a matching DPoP header should be required; no header should be required otherwise. | Protected resources should require a matching DPoP header if the supplied access token is DPoP-bound; should ignore the header otherwise. |
+| Disabled | DPoP headers should be ignored; issued tokens shouldn’t contain JWK thumbprints. | DPoP headers and token JWK thumbprints should be ignored. |
+
```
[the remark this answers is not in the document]
tnorimat: By looking at [DPoP's draft](https://tools.ietf.org/id/draft-ietf-oauth-dpop-02.html#name-dpop-access-token-request), if we make refresh token DPoP-bound, we need to use the same public key whenever refreshing tokens. Therefore, we need to use and keep the same public key until the first issued original refresh token expires or is revoked. Considering that the lifetime of the refresh token is much longer than the access token, a SPA as the client without its secure storage needs to keep it longer, which might be not good from security perspective.

design/dpop.md
```diff
@@ -0,0 +1,259 @@
...
+* **JIRA**: TBD
+
+## Motivation
+
+DPoP is a mechanism to prevent illegal API calls from succeeding only with a stolen access token. This is done via augmenting the calls to OAuth 2.0 endpoints as well as resource servers with a special DPoP header that proves possession of a private key, and binding the issued tokens to that key. Please refer to the Resources section for more info on DPoP.    
+
+## Implementation Details
+
+DPoP is a cross-cutting concept that would affect multiple Keycloak components, namely core libraries, OAuth/OIDC endpoints, adapters, models and admin UI.
+
```
tnorimat, line 13: It might be good if the adapters takes what roles. I infer that those are used as a resource server or its reverse proxy.

design/dpop.md
```diff
@@ -0,0 +1,259 @@
...
+_([5.5. DPoP Access Token Request][6])_
+
+> :question: As DPoP is primarily meant to protect tokens issued to SPAs, should we also support DPoP for:
+>	
+> * password grant 
+> * client credentials grant (DPoP-enabled service accounts?)
+> * token exchange
+> * extension grants (with the future Grant Type SPI)
+> * UMA permissions grant?
+
```
tnorimat, line 56: IMO, it might be preferable that this document only treats basic authorization code grant and token refresh. Other grant could be realized by follow-up tasks after this document is merged and corresponding PR is also merged.
The reason why is that if this document treats all grants shown here, the corresponding PR's code volume would become huge, which makes the keycloak development team hard to review it and might takes a lot of time for this PR to be merged.

design/dpop.md
```diff
@@ -0,0 +1,259 @@
...
+Files affected:
+
+*   adapters/oidc/js/src/main/resources/keycloak.d.ts
+    
+
+#### Other Adapters
+
+Support in Node.JS adapter, mod\_authn\_oidc and Keycloak Gatekeeper could be also considered.
+
```
tnorimat, line 200: [As you may know](https://www.keycloak.org/2020/08/sunsetting-louketo-project.adoc), Keycloak Gatekeeper has already been EOL.

design/dpop.md
```diff
@@ -0,0 +1,259 @@
...
+Affected documents:
+
+*   Securing Applications and Services Guide
+    
+## Open Questions
+
+1.  Which OAuth grant types should be DPoP-enabled?
+2.  Should we support Java adapters' redirect mode (for monolithic webapps)?
+3.  How do we implement adapter-side DPoP proof replay protection in a clustered environment?
+4.  Should we generate a DPoP keypair per session, or should it be persisted in the user agent (or both)?
```
tnorimat, line 242: Basically, generating key pair per session might be good from security perspective. However, if the client can use some secure storage, per user agent might be OK.

tnorimat:
Hello @dteleguin, I've added some comments. I'm happy if you could check them.

design/dpop.md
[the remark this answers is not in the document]
dteleguin: Good point, will add a section on refresh tokens. The above should be relevant to offline tokens as well, since those are basically "immortal" refresh tokens. Hence, the keypair should be persisted for the lifetime of the offline token.

Also, AFAIK Keycloak always issues a new refresh token upon refresh ([link](https://github.com/keycloak/keycloak/blame/master/services/src/main/java/org/keycloak/protocol/oidc/TokenManager.java#L343)), so this will result in a chain of refresh tokens bound to the same DPoP keypair; so the keypair should be kept for the duration of the entire session.

design/dpop.md
dteleguin: Agreed, will move other grant types to milestones.

design/dpop.md
dteleguin: Oh indeed. Though OAuth2 Proxy is meant to be the successor, I'm not sure if DPoP would be relevant for that type of adapters, so will remove the mentioning for now.

design/dpop.md
dteleguin: Agreed. We can start with a transient (per-session) keypair in JavaScript adapter, and add WebCrypto-based key storage later (will move that to milestones).

design/dpop.md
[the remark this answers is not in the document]
tnorimat: Yes, I think so.
It might be better that we have a chance to change the key when we refresh the token. However, I have not found how to do it in [the DPoP specification](https://tools.ietf.org/id/draft-ietf-oauth-dpop-02.html#section-5). 

This DPoP specification treats the following things individually.
- issue holder-of-key bound token (by AS)
- verify holder-of-key bound token (by RS)

and does not treat them simultaneously (eg. refresh tokens)
- verify holder-of-key bound old token and issue holder-of-key bound new token (by AS)
- 
which means that we only use the same key for both old and new token. Also it leads to using the same key during the refresh token (also offline token in OIDC) is active. Refreshing keys is good from security perspective but we cannot do it.
 
It might be better to have the way of using separate key for each old and new token and moreover be specified by DPoP specification.

design/dpop.md
tnorimat: To say in more detail, there are several keycloak adapters and some are for client applications while others are for reverse proxy which can do DPoP related tasks on behalf of Resource Server. I would like to know whether this design document considers only client applications or both.

design/dpop.md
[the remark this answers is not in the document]
dteleguin: Excellent point - the DPoP spec currently doesn't mention any means for key rotation; I think we should approach the authors of the DPoP spec with this. The solution might be as simple as providing two DPoP proofs in a single header (for the special case of token refresh):

```
POST /token HTTP/1.1
Host: server.example.com
Content-Type: application/x-www-form-urlencoded;charset=UTF-8
DPoP: eyJ0eXAiO... eyJ0eXAiO...

grant_type=refresh_token
&refresh_token=Q..Zkm29lexi8VnWg2zPW1x-tgGad0Ibc3s3EwM_Ni4-g
```

This would allow the server to use the first DPoP proof for verification, and the second one for the issuance of a new token set. Or perhaps those could be two different headers.

design/dpop.md
dteleguin: Thanks for clarifications. I think we'd focus on JavaScript and Java adapters first, since the primary use case for DPoP is SPAs.

tnorimat: Thank you for your clarification.

dteleguin: In regards to reverse proxy type adapters, I think supporting DPoP could be low priority since access tokens normally are not exposed to the user agent (though they might be, via a custom header).

Also supporting DPoP for proxies might be relevant if there is a requirement to connect the proxy adapter to a client that enforces DPoP (i.e. doesn't support Bearer and DPoP auth schemes simultaneously), but that should be a rare case.

Therefore, I think it should be safe to de-prioritize proxy adapter support.

design/dpop.md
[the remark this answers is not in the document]
tnorimat: According to https://tools.ietf.org/html/draft-ietf-oauth-dpop-03#section-5 , 

> When an authorization server supporting DPoP issues a refresh token to a public client that presents a valid DPoP proof at the token endpoint, the refresh token MUST be bound to the respective public key. 
> The binding MUST be validated when the refresh token is later presented to get new access tokens.  As a result, such a client MUST present a DPoP proof for the same key that was used to obtain the refresh token each time that refresh token is used to obtain a new access token.  

Therefore, by following this paragraph, we need to use the same key that was used to get the first access and refresh token whenever we refresh tokens. As a result, all access and refresh tokens are bound with the same key until refreshing tokens is invalidated (ex. expiration of the first refresh token, revoking it).

design/dpop.md
tnorimat: Thank you for your clarification.

design/dpop.md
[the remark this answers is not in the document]
mposolda: Yes, this missing support for key rotation looks like a limitation of DPoP specification...

One possible workaround for this is to make sure that client does full redirect for the SSO re-authentication instead of refresh tokens. For example we can document for the javascript adapter (and maybe other parts of the documentation) that client is encouraged to do full SSO redirect after some time (EG. every 30 minutes) or after some count of refreshes (EG. 5). SSO redirect will use same userSession and hence there is not additional data needed on the server side.

This may not be ideal for some kind of applications as SSO redirect can cause the lose of some state of the javascript application in the browser.  And DPoP with refresh token is intended to be used primarily for the public clients according to the specification and hence primarily javascript applications. But at least we can document that such possibility exists as it can be used at least by some applications. WDYT?

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+   the client MUST provide a valid DPoP proof JWT in a "DPoP" header
+   when making an access token request to the authorization server's
+   token endpoint.  This is applicable for all access token requests
+   regardless of grant type (including, for example, the common
+   "authorization\_code" and "refresh\_token" grant types but also
+   extension grants such as the JWT authorization grant \[RFC7523\]).
+
+_([5.5. DPoP Access Token Request][6])_
+
+First and foremost, the `authorization_code` and `refresh_token` grant types must be supported; support for other grant types (ROPC, client credentials, token exchange, extension grants) could be added later.
```
mposolda, line 58: Note for refresh tokens: According to the specification `Refresh tokens issued to confidential clients (those having established authentication credentials with the authorization server) are not bound to the DPoP proof public key because they are already sender-constrained with a different existing mechanism.` . So it seems to me that we do not need to bind refresh tokens to DPoP for the confidential clients?

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Motivation
+
+DPoP is a mechanism to prevent illegal API calls from succeeding only with a stolen access token. This is done via augmenting the calls to OAuth 2.0 endpoints as well as resource servers with a special DPoP header that proves possession of a private key, and binding the issued tokens to that key. Please refer to the Resources section for more info on DPoP.    
+
+## Implementation Details
+
+DPoP is a cross-cutting concept that would affect multiple Keycloak components, namely core libraries, OAuth/OIDC endpoints, adapters, models and admin UI.
+
+It is suggested that DPoP should be set up on a per-client basis, with the three options available that should be interpreted as follows:
```
mposolda, line 14: One consideration is, whether to use client-specific configuration options or rather rely on client policies? It will be nice if we can move much of the client-specific configuration options from clients and drive them with the use of client policies. 

Especially for DPoP it can be nice to do that due the fact that DPoP specification does not seem to introduce any client-specific metadata for client registration?

@tnorimat WDYT?

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+| Optional | _Token Endpoint_: if DPoP header is present, the returned token should be DPoP-bound and should contain a `cnf` claim with a JWK thumbprint; should be a regular token otherwise  <br/><br/>_UserInfo endpoint_: if the access token is DPoP-bound, a matching DPoP header should be required; no header should be required otherwise. | Protected resources should require a matching DPoP header if the supplied access token is DPoP-bound; should ignore the header otherwise. |
+| Disabled | DPoP headers should be ignored; issued tokens shouldn’t contain JWK thumbprints. | DPoP headers and token JWK thumbprints should be ignored. |
+
+Other DPoP configuration options should include:
+
+| **Option** | **Description** |
+| --- | --- |
+| DPoP Proof Lifetime | Time window in which the respective DPoP proof JWT would be accepted |
+| DPoP Allowed Clock Skew | To accommodate for clock offsets, the server MAY accept DPoP proofs that carry an `iat` time in the reasonably near future (e.g., a few seconds in the future). |
```
mposolda, line 27: Is this config option really needed? Just wondering that ideal would be to avoid many configuration options as each configuration option is adding some complexity (And Keycloak client already has really big amount of configuration options). For example JWTClientAuthenticator does not check that "iat" from the token is in the past (see https://github.com/keycloak/keycloak/blob/15.0.2/services/src/main/java/org/keycloak/authentication/authenticators/client/JWTClientAuthenticator.java) and hence it seems to me to have same behaviour for DPoP as well?

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+* Token Introspection
+* Token Revocation
+
+The following endpoints use cookie authentication:
+
+* Check Session iframe
+
+Therefore, no other endpoints should be affected.
+
+#### Client Registration
```
mposolda, line 131: Are there any client specific metadata for DPoP? I am not seeing any in the https://tools.ietf.org/id/draft-ietf-oauth-dpop-03.html specification, but maybe I am missing something?

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+* DPoP proof lifetime (integer)
+* DPoP clock skew (integer)
+
+Those could be backed by client attributes, hence no changes to the client model and DB schema.
+
+Classes/methods affected:
+
+* org.keycloak.protocol.oidc.OIDCAdvancedConfigWrapper
+
+### Adapters
```
mposolda, line 164: TBH, I would leave adapters as the lowest priority unless there is an explicit request for adding DPoP support to them. Do you have a need for adapters support for your deployment?

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+
+Classes/methods affected:
+
+*   org.keycloak.adapters.\*
+
+##### Application Clustering
+
+As per the spec,
+
+> Servers SHOULD store, in the context of the request URI, the `jti` value of each DPoP proof for the time window in which the respective DPoP proof JWT would be accepted and decline HTTP requests to the same URI for which the `jti` value has been seen before.
```
mposolda, line 190: As you pointed, this is quite challenging to support especially in the cluster environment. I would personally not support this requirement for now (Specification says "SHOULD") unless there is very strong requirement of it. I see there is some security gain, however the price to pay store additional data on the resource servers side is the challenge (especially for the clustering environment).

mposolda (changes requested):
@dteleguin Thanks for the specification and sorry for the long time to review...

Added few inline points to the specification. I am marking this as "Request Changes" until we resolve the points.

design/dpop.md
tnorimat: @mposolda Yes, as you pointed out, the current [DPoP draft(v3)](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop-03) does not introduce any client metadata. Therefore, it might be better to rely on client policies and not introduce a new client setting for DPoP.

tnorimat:
@dteleguin , @mposolda 
The one of the discussion points is to whether we add client settings for DPoP or not.

By adding client settings, we can configure DPoP operation per client, but we still have vast number of client settings so that keycloak admin needs to pay costs for maintaining these client settings value being valid.

By using client policies, we can enforce some operations (mode : required/optional/disabled, DPoP Proof Lifetime, DPoP Allowed Clock Skew) on several clients, which makes us not to add additional client settings.

My vote is to use client policies.  I am familiar with client policies so that I can help for implementing DPoP with client policies.

design/dpop.md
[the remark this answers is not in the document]
dteleguin: @mposolda agree on that; we can mention this as "best practices" in the JS adapter docs.

Just in case, here is the stance on DPoP key rotation from from the spec co-author:

> (...) the general rough consensus thinking has been
that allowing for DPoP key rotation for refresh tokens and public clients
(the only case where it's relevant) didn't add enough value to justify the
added complexity. It doesn't help with the threat model for in-browser
applications. And mobile applications have really good options for key
storage - to the point that the kind of event that might compromise a DPoP
key would involve a lot more than key rotation to cleanup from.

[(ML thread)](https://mailarchive.ietf.org/arch/msg/oauth/65Aip95ixdP8PplzFXuIfREERvc/)

design/dpop.md
dteleguin: Agreed on that; will incorporate into the prototype implementation.

design/dpop.md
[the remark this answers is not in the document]
mposolda: @dteleguin Thanks for the update and referencing this discussion!

VinodAnandan:
New draft: https://www.ietf.org/rfcdiff?difftype=--hwdiff&url2=draft-ietf-oauth-dpop-04.txt

design/dpop.md
dteleguin: > For example JWTClientAuthenticator does not check that "iat" from the token is in the past

Did you perhaps mean "in the future"? The only check that is performed against `iat` is this:

```java
            // KEYCLOAK-2986
            int currentTime = Time.currentTime();
            if (token.getExpiration() == 0 && token.getIssuedAt() + 10 < currentTime) {
                throw new RuntimeException("Token is not active");
            }
```

That means, we're accepting `iat` values that are no more than 10 seconds into the past + values set in indefinite future. 

Unfortunately, this might not work for DPoP proofs. Let's revisit section 10.1 "DPoP Proof Replay":

> If an adversary is able to get hold of a DPoP proof JWT, the adversary could replay that token at the same endpoint (the HTTP endpoint and method are enforced via the respective claims in the JWTs). To prevent this, servers MUST only accept DPoP proofs for a limited time window after their iat time, preferably only for a relatively brief period (on the order of a few seconds).

> Servers SHOULD store, in the context of the request URI, the `jti` value of each DPoP proof for the time window in which the respective DPoP proof JWT would be accepted and decline HTTP requests to the same URI for which the `jti` value has been seen before. In order to guard against memory exhaustion attacks a server SHOULD reject DPoP proof JWTs with unnecessarily large jti values or store only a hash thereof.

> Note: To accommodate for clock offsets, the server MAY accept DPoP proofs that carry an `iat` time in the reasonably near future (e.g., a few seconds in the future). Because clock skews between servers and clients may be large, servers may choose to limit DPoP proof lifetimes by using server-provided nonce values rather than clock times, yielding intended results even in the face of arbitrarily large clock skews.

Key points are:
- in order to prevent DPoP proof replays, we must temporarily store the `jti` values;
- we should keep the `jti` values in memory only for the limited time window (aka DPoP proof lifetime), otherwise we'll quickly exhaust memory;
- DPoP proofs have neither `nbf` nor `exp` claims, only `iat` which is provided by the client; hence, we should enforce limited window of validity based on client-supplied `iat` only;
- clock skews can be significant (and could even exceed proof lifetimes, see [this](https://github.com/danielfett/draft-dpop/issues/67)), but
- we cannot allow `iat`s set in indefinite future, since the server will have to store `jti` until (`iat` + lifetime) is reached, which creates possibility of conducting memory exhaustion attacks. Hence, allowed clock skew has to be finite.

That said, we should have two server-side parameters, one for DPoP proof lifetime and another one for allowed clock skew. I decided to make them configurable, in order to better suit particular setups and security requirements, similar to what we have in another parts of Keycloak (configurable token lifetimes; configurable clock skews for adapters and IdPs).

The recent version of the DPoP draft offers an alternative mechanism of limiting DPoP proof lifetimes, which is based on server-supplied nonces rather than clock times (see [8. Authorization Server-Provided Nonce](https://tools.ietf.org/id/draft-ietf-oauth-dpop-04.html#section-8) and [9. Resource Server-Provided Nonce](https://tools.ietf.org/id/draft-ietf-oauth-dpop-04.html#section-9)) and is said to "yield intended results even in the face of arbitrarily large clock skews". However, this method is more complex and won't be featured in the initial PR. And even with this method we'll have to somehow configure lifetimes of server-supplied nonces.

design/dpop.md
dteleguin: Agreed; no, we do not have any requirements for the adapters. Will reflect that in the doc.

design/dpop.md
dteleguin: Agreed.

mrpatrickpilch:
Are there still plans to implement this? 

I think it would be a great security addition, since mobile devices' secure enclaves can generate non exportable private keys and enable very effective device binding.

dteleguin:
@mrpatrickpilch yes, this is a work in progress now, and has very high priority. You can track the progress under the [draft PR](https://github.com/keycloak/keycloak/pull/8895).

It is fairly usable now, bar some functions like client metadata and support for DPoP at UserInfo. If you're ready to engage in some testing and provide feedback, that would be much appreciated!

design/dpop.md
```diff
@@ -0,0 +1,308 @@
...
+Classes/methods affected:
+
+* org.keycloak.representations.AccessToken.CertConf
+	* rename to Conf or Confirmation; introduce `keyThumbprint` property
+
+#### Client
+
+There are the following DPoP-related client settings:
+
+* DPoP mode (enabled/optional/disabled)
```
tnorimat, line 154: To simplify the implementation, it is only needed to add one client settings:
- dpop_bound_access_tokens (true/false)
dpop_bound_access_tokens is a client metadata defined in the DPoP specification.
The boolean value is easy to handle compared with ternary (enabled/optional/disabled).

As for DPoP proof lifetime and clock skew, it might be better to use default values for all clients. It might be better not to add new client settings because keycloak has already a lot of such settings and these settings are difficult to be managed.

If we add client settings at this time and we add some method (e.g., client policies) to remove necessities of the settings afterwards,  we cannnot delete these settings because we need to keep backward compatibility.

Therefore, at this time, it might be preferable to use default values only for DPoP proof lifetime and clock skew, and add some mechanism to change these values per client or set of clients without adding client settings afterwards.

tnorimat (commented):
@dteleguin @mposolda I re-reviewed it and add one comment about client settings. Could you check it?

PR #258, opened 2021-03-27
KEYCLOAK-2045 Validator SPI Design Doc

thomasdarimont:


velias:
Thanks @thomasdarimont  for this design doc. It thoroughly summarizes discussions held in UserProfile SPI working group and their conclusions. I believe proposed Validation SPI is simple and flexible enough base to cover rich variety of validation needs in keycloak. Mainly validations configured in realms by admins necessary for UserProfile SPI prevent us from using code annotation oriented "Bean Validation API". 
I can see three areas of future extensions once this base is settled and merged:
* ability to simply construct validation definitions for more complex objects (FormData and REST API input data Representatins) in code and run them against these objects in one simple call to make validations for these basic cases as simple as possible
* integration of validation results/error messages from Validation SPI into forms (freemarker templates, authenticator forms) with i18n support to make templates development easier/seamless
* integration of validation results/error messages into REST API responses framework (with i18n binding?) to make REST API development easier but also standardize representation of validation errors in REST API responses

vmuzikar:
Would it make sense to give this a shout-out on keycloak-dev? IMHO this is a pretty important SPI and we should make sure it has a proper "visibility" before setting it in stone. Or has this been already discussed (e.g. as part of the User Profile WG)? @hmlnarik

thomasdarimont:
@vmuzikar just posted a message to the keycloak-dev group: https://groups.google.com/g/keycloak-dev/c/9CfGDGngeeU

vmuzikar (commented):
@thomasdarimont Thank you for the design proposal. I really like it! It's elegant yet not overcomplicated. Just a few questions/thoughts.
* How would more complex validations involving multiple fields work? Something like validate that street number and ZIP code are a valid address, if street number and ZIP code passed some other validations (like have they a correct format etc.)?
* As already mentioned by @velias, I think we should have a built-in i18n support.
* We might want to be able to validate whole representations, see e.g. one of the [current validation "frameworks"](https://github.com/keycloak/keycloak/blob/01be601dbdd77822827de173e34180d9322db85c/services/src/main/java/org/keycloak/validation/DefaultClientValidationProvider.java).
* Just a thought (probably out of scope of this SPI). Do we want to do all validations just at the server side? Don't we want to have some overlap (not sure how exactly) to e.g. JS so (some) validations could be done e.g. in new Admin Console? @ssilvert

thomasdarimont:
@vmuzikar Thanks for your feedback!

1) I just added an example for performing [nested validation](https://github.com/keycloak/keycloak/pull/7887/files#diff-e8a31e5e873f1c5826f3c2ce581056a85457d21bc60504d06873410ef36f9bcbR209). Is this something you had in mind?
2) We already have the (i18n) message + (optional) messageParameters, rendering of error messages could be done before the errors are sent to the UI.
3) This can easily be done with custom validators and nested validations, see [nested validation](https://github.com/keycloak/keycloak/pull/7887/files#diff-e8a31e5e873f1c5826f3c2ce581056a85457d21bc60504d06873410ef36f9bcbR209)
4) At least for the user-profile validation config it should be possible to use the validation metadata to drive a slim client side validation in JS as well.

pedroigor:
@vmuzikar @hmlnarik Hope you guys can join us in this effort.

The goal is a general-purpose validation API, where the changes proposed by @thomasdarimont are the baseline for being used not only in user profile but other parts of Keycloak.

@thomasdarimont We should probably include here what we discussed about having either the validation utility class or the possibility to integrate this validation to the `KeycloakSession` so that we can remove a lot of redundant and boilerplate code too.

thomasdarimont:
@pedroigor regarding the `KeycloakSession` integration.

To ease `Validator` lookups, I thought about having a `KeycloakSession#validator(String validatorId)` method, which delegates to the `ValidatorLookup#validator(String validatorId)`.

This would enable users to write code like this:
```java
session.validator("address").validate(myAddress).isValid();
```

Currently user have to write:
```java
ValidatorLookup.validator(session, "address").validate(myAddress).isValid();
```

pedroigor:
> 
> This would enable users to write code like this:
> 
> ```java
> session.validator("address").validate(myAddress).isValid();
> ```

With the introduction of the user profile, users won't need to write this code at all, right?

I mean, all that is going to be provided OOTB by the provider.

I do see value though for things lile:

```
session.validators().isNullOrEmpty(string);
session.validators().isEnabled(realm);
session.validators().isSslRequired(realm);
```

In other words, methods that could remove a lot of boilerplate code that is spread all over the places and related to validation.

Another example is completely replace the very specific code for client validation (ClientValidationProvider), maybe with something like:

```
session.validators(client).hasValidUrls().
```

The main point being, the entry point you are proposing would be used and re-used in every single place for whatever domain model.

pedroigor:
> * Just a thought (probably out of scope of this SPI). Do we want to do all validations just at the server side? Don't we want to have some overlap (not sure how exactly) to e.g. JS so (some) validations could be done e.g. in new Admin Console? @ssilvert

Yes, it is out of scope.

Also, the user profile also uses metadata sent to clients to give them info about how things should be rendered and validated. Together with dynamic forms/templates, UI could be generated dynamically.

I don't see why you want these type of dynamic in admin console. IMO, just complicates things. In fact, even for user profile this is also a complicating factor.

thomasdarimont:
> This would enable users to write code like this:
>> session.validator("address").validate(myAddress).isValid();

> With the introduction of the user profile, users won't need to write this code at all, right?

Actually, users might want to use this to validate their Provider configurations (Protocol Mappers, Federation Providers etc.).

> session.validators()

Could return a `ValidatorLookup` instance -> we could call that class `Validators`. `Validators` could provide a bunch of simple validation methods as you suggested, in addition to the possibility for `Validator` lookups by provider id.

This would then enable the use of `session.validators().validateNotEmpty(...)` etc. for simple validations and 
`session.validators().validator(...).validate(...)` for the dynamic `Validator` lookups needed by the user profile.

> `session.validators(client).hasValidUrls()`
This would imply that we'd need a type based `Validator` mapping/lookup:
- `session#validators(Object o)` -> we could then try to find a `Validator` for o.getClass().
- `session#validators(String validatorId)` -> we perform a direct lookup for a `Validator` with that `providerId`

If you want type specific checks like `session.validators(client).hasValidUrls()` or `session.validators(user).hasValidEmail()` then you need to have overloaded methods that return the proper `Validator`.

- `session#validators(UserModel u) returns UserModelValidator`
- `session#validators(RealmModel r) returns RealmModelValidator`
- `session#validators(ClientModel c) returns RealmModelValidator`
- `session#validators(String validatorId) returns Validator` -> we perform a direct lookup for a `Validator` with that `providerId`

This should be possible to add on top of the current implementation.

thomasdarimont:
```java
public interface ClientModelValidator {

    boolean hasValidUrls();

    // other ClientModel specific validations
}
```

```java
public class Validators {
 	...
	private final KeycloakSession session;

    public Validators(KeycloakSession session) {
        this.session = session;
    }
	...

    public ClientModelValidator validator(ClientModel client) {
        return new ClientModelValidator() {
            @Override
            public boolean hasValidUrls() {
                ValidationContext context = new ValidationContext();
                for (String uri : client.getRedirectUris()) {
                    // static facade method uri() returns UriValidator.INSTANCE
                    uri().validate(uri, "redirectUri", context);
                }
                return context.isValid();
            }
        };
    }
	
```

Code which wants to use the validation can then do this:
```java
	session.validators().validator(client).hasValidUrls();
```

in `KeycloakSession` we could then have:

```java
class DefaultKeycloakSession {
  ...

  public Validators validators() {
	return new Validators(this);
  }
}
```

thomasdarimont:
I updated the Validation API proposal with `Validators` as the combination of `ValidatorLookup` and `BuiltinValidators` as described above. `Validators` is now a facade and serves as the central goto-place for all things regarding validation.

jbman:
Hi @thomasdarimont, I did take a look at the implementation as we want to use validation SPI for user profile definitions in the future. Looks like a good approach. 
Is the plan to support validator configuration key-values managed at Keycloak later or will it the SPI implementation be responsible to provide the config object? Validator configuration maps managed at Keycloak for user or for other representations would be a nice.

Also added detail questions inline.

thomasdarimont:
@jbman thanks for your feedback!

The idea is that Keycloak will ship with a set of built-in `Validator` implementations and allows for custom user-provided `Validator` implementations. Some of those `Validator`s support a parameterized validation. This parameterization is provided via a `ValidatorConfig` object that is passed to the `Validator#validate(...)` method as a parameter. `ValidatorConfig` objects can be build from simple Maps via the `ValidatorConfig#configFromMap` method. The config can be provided via the `validations` configuration of a user profile definition, e.g.:
```json
...
"email": {
  validations: [ 
    { validator: "email" }, 
    { validator: "pattern", config: { pattern: ".*@acme.com" }}
  ]
}
...
```

This configuration will then be used when a concrete profile based on the above definition is validated.

The test [public void validateValidatorConfigViaValidatorFactory()](https://github.com/keycloak/keycloak/pull/7887/files#diff-e8a31e5e873f1c5826f3c2ce581056a85457d21bc60504d06873410ef36f9bcbR227) shows how this may look like. Note that most of the time users will not interact with this API themselves, since the Keycloak (UserProfile) infrastructure will take care of that.

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+1. Promote Validation SPI to `server-spi` as an official API and expose validator lookup via KeycloakSession
+
+### Open Questions
+1. Should users be able to override built-in Valdators?
+1. How should the error message keys look like?
+1. Do we need a concept like a `validation hook` to allow running user provided Validators before/after/instead of built-in validators?
+1. Should ValidationErrors be Comparable? Do we need concepts like `order` or `priority`?
+
+### Further Work
+1. Add support to report `ValidationWarning`s next to `ValidationError`s to report problems, that don't fail the validation. A generic base type could be `ValidationProblem`.
```
jbman, line 292: It doesn't need to be the validators responsibility to decide if sth. is a warning or an error. Same validator could be called with different configurations => We only need one ValidationError but one validatar config for warnings and one for errors.
Example: A phone number could be validated with min length 3 max mlength 20 to produce warning on a warning on length < 3. This is just a hint that some totally wrong value might be put into the field.
A second validation uses just max max length = 100. Because this is a strict limit.

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+1. Email - Checks if the given value is a valid email address
+
+### Integration plan
+
+1. Prepare Validation SPI in `server-spi-private` in the `org.keycloak.validation` package and consolidate the other Validation APIs to use the new Validation SPI. Note, that most of the validation code can be directly replaced.
+1. Adapt User-Profile support to use the new Validation SPI
+1. Promote Validation SPI to `server-spi` as an official API and expose validator lookup via KeycloakSession
+
+### Open Questions
+1. Should users be able to override built-in Valdators?
```
jbman, line 286: IMHOwWherever a validator is used for representation validation it shouldbe possible to specify a custom one to be used (which could make use of built-ins low-level validators internally). Build-in low-level validators don't need to be overrideable.

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+
+Some usage examples can be found here in the [ValidatorTest](https://github.com/thomasdarimont/keycloak/blob/issue/KEYCLOAK-2045-Simple-Validation-SPI/server-spi-private/src/test/java/org/keycloak/validate/ValidatorTest.java).
+
+### Core API
+The proposal focus around the following types:
+1. `Validator`: `Provider` interface for validation mechanics.
+1. `ValidationContext`: Holds state of a sequence of validations.
+1. `ValidationResult`: Denotes the outcome of a validation.
+1. `ValidatorConfig`: Denotes the configuration for a Validator. A typed wrapper around a `Map<String,Object>`.
+1. `ValidationError`: Represents an error found during validation.
```
hmlnarik, line 112: I `ValidationError` any different ot `ValidationResult`?

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+
+- hibernate-validator is currently not used in Keycloak, thus we'd need to add a dependency for:
+```xml
+<dependency>
+    <groupId>org.hibernate</groupId>
+    <artifactId>hibernate-validator</artifactId>
+    <version>6.0.12.Final</version>
+</dependency>
+```
+- No direct support for validation (Constraint) lookups by name. The user-profile support needs a way to lookup validations by name.
```
hmlnarik, line 76: The support for lookup by name is as easy as providing a simple map from name to validator class - just the same as (`ValidatorLookup`) proposed in this PR.

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+1. `ValidatorLookup`: Helper class to provide lookups for built-in and user provided validations.
+
+### Support API
+The following types provide the integration with the Keycloak infrastructure:
+1. `ValidatorFactory`: `ProviderFactory` interface for custom validation contributions.
+1. `ValidatorSPI`: Provides the `validator` SPI.
+1. `CompactValidator`: Convenience class for built-in validations.
+1. `BuiltinValidators`: Denotes a registry for internal built-in validations.
+
+The proposed package name is `org.keycloak.validation`. Note the current PR [KEYCLOAK-2045 Simple Validation API](https://github.com/keycloak/keycloak/pull/7887) uses the package name `org.keycloak.validate` to provide a transition path for the current API.
```
hmlnarik, line 122: This list needs big elaboration. How do these classes work together? How does it compare to to JSR 380?

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+
+An earlier attempt at a unified validation SPI can be found here [KEYCLOAK-2045 Add support for flexible Validation](https://github.com/keycloak/keycloak/pull/7324) and was discussed on the [keycloak-dev mailing list](https://groups.google.com/g/keycloak-dev/c/-XjQu7rn56s).
+
+After some discussion, it became apparent that this proposal provides a very comprehensive validation framework that 
+offers numerous functionalities that go beyond the requirements of user profile validation and other validations in Keycloak.
+
+However, the API was very complex and not easy to understand. Therefore, it was agreed that @thomasdarimont would revise 
+the proposal again and present it in a smaller form. This has resulted in the new "Simple Validation SPI" proposal.
+
+## Proposed Validation SPI
```
hmlnarik, line 98: I am missing the proposal here. PoC is not sufficient to be a proposal. This document needs to describe what is proposed.

hmlnarik (changes requested):
Thank you @thomasdarimont for the proposal.

I think this is a good start.

Few general questions here:
- How are the validators expected to be perform validations that depend on multiple fields?
- How are the validators expected to be wired into the flow to ensure the validation of the data? Is it by putting a call to validator to each REST endpoint? Some other mechanism? Please elaborate.

Please also see inline. Note that the intention here is to find a good solution to validation so that is may be used long term. I am not pushing for JSR 303, 380 or any other standard if another way would prove beneficial and is properly justified. At the same time, I am so far not convinced that this is the case of this proposal now. I am open to changing this opinion as the proposal evolves.

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Design Principles
+
+The design of the new validation SPI should follow the following design principles:
+
+- Simplicity
+The API should have a small API surface and be easy to use.
+
+- API Granularity
+The API should support low-level validations.
```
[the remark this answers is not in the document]
thomasdarimont: Examples for "low-level" validations are: 
- `email`: check if the given string is a well-formed email address
- `notEmpty`: check if the given string (or collection...) is not empty
- `length`: check if the given string has a certain length (min, max can be configured via `ValidationConfig`)
- `pattern`: check if the given string matches a given pattern
- `url`: check if the given string is a well-formed URL.

Those validations are applied to an single value (an email address, a username, URL of a company website) etc. They are simple and might allow some configuration such as pattern, url/uri requirements, min/max length etc. User's will most likely use "low-level" validations as is. 
Keycloak should provide some simple validations out of the box.
On the other hand we have "High-level" validations, which might denote complex multi-field validations that are composed of a set of "low-level" validations, e.g. to validate a `UserModel` or a `ClientModel` as a whole.

Alternative terms for `low-level` and `high-level` validations might be:
- "simple" and "complex" validations.
- "singular" and "compound" validations.

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+- Extensibility
+There should be support for built-in as well as user provided validations.
+The new SPI should be easy to extend, most custom validations should only need one implementation class. 
+Developers should be able to provide reusable custom validation logic components.
+
+- Versatility
+Validations should be flexible and should support validatation of simple values and complex objects.
+It should be possible to parameterize validation executions. Additionally there should be a way to 
+check if the parameterization for a validation is valid.
+Validations should be composable and it should be possible to access other validations from with a validation.
```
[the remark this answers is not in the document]
thomasdarimont: Thanks!

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+## Considered Options
+
+### Bean Validation API
+
+The JSR 380 Bean Validation 2.0 is a mature and rich validation framework that is extensively used in Java applications.
+
+- Optimized for validation of complex Java bean objects based on declarative configuration rather than single values.
+
+Allthough, value validations are possible via `javax.validation.Validator#validateValue`, they require a bean class 
+with a property with the annotated constraints.
```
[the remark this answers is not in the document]
thomasdarimont: That's nice and those programmatic constraint definitions can be created from a `UserProfile` config, but how does this work with the dynamic nature of `UserProfile`? There are no explicit fields or methods that expose the user profile properties, since it's effectively backed by a `Map` of attributes. Is there a way to plug-in your own property resolvers?

design/validation-spi.md
```diff
@@ -0,0 +1,293 @@
...
+- No easy way to dynamically configure constraints programatically. The user-profile support needs a way to configure validation dynamically.
+- Definition of [custom validations is quite involved](https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#validator-customconstraints) - you need 2 classes (annotation + constraint) + error message + META-INF service manifest.
+
+All this was considered and evaluated in a small prototype, but then it was concluded that it makes more sense to go 
+with a custom implementation that provides a narrow API. Additionally custom extensions should be easier and provide 
+a better integration with the existing Keycloak infrastructure. Further more dynamic validation logic lookups 
+and configuration should be possible and easy to do. 
+
+Note that it is still possible to provide a validation/validator implementation that leverages bean validation. 
+Also, bean-validation still works quite well for declarative validation in JAX-RS endpoints.
```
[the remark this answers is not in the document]
thomasdarimont: What I thought, was that support for declarative bean validation might make sense in the context of JAX-RS resources to express constraints for things like query parameters, form parameters etc.

But I agree, that validation should be done on the model level in any case.

design/validation-spi.md
thomasdarimont: A `ValidationError` denotes a concrete error found during validation that was reported by a `Validator`. `ValidationResult` contains a collection of all the errors found during validation.

design/validation-spi.md
thomasdarimont: I agree with you that you can build this mapping yourself.  I was just looking for a native way to do the lookups.

design/validation-spi.md
thomasdarimont: The proposed Validation SPI is simpler than JSR 380 and deliberately provides less features. It it works without reflection and does not use annotations. The extension model is aligned with Keycloak's extension mechanism.

design/validation-spi.md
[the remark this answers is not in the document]
vmuzikar: +1 for putting validations on the model level. But how do we let the higher levels know that validation failed. E.g. the REST endpoints won't get the validation result. Should we have some dedicated `ValidationException`? If so, it should be reflected in the design document.

design/validation-spi.md
thomasdarimont: Okay, I think this needs some clarification. The Proposed Validation SPI section is more or less an overview and explanation of the [existing Validation SPI PoC](https://github.com/keycloak/keycloak/pull/7887).

I see if I can summarize this a bit here.

vmuzikar:
> 1. I just added an example for performing [nested validation](https://github.com/keycloak/keycloak/pull/7887/files#diff-e8a31e5e873f1c5826f3c2ce581056a85457d21bc60504d06873410ef36f9bcbR209) -> look for `public void nestedValidation()`. Is this something you had in mind?

Yes, thanks. The last think I'm wondering about is how to handle errors for compound fields. We want to add a single error message covering multiple fields (e.g. address – street, city, ZIP – is incorrect). Should we have some kind of `CompoundValidationError`?

> 2. We already have the (i18n) message + (optional) messageParameters, rendering of error messages could be done before the errors are sent to the UI -> look for `public void formatError()`.

Yes but what I had in mind was a built-in support for i18n for convenience (something like [here](https://github.com/keycloak/keycloak/blob/01be601dbdd77822827de173e34180d9322db85c/server-spi-private/src/main/java/org/keycloak/validation/ValidationError.java#L61-L68)). And if the validations will be part of the `KeycloakSession`, we might not even need the explicit `messagesBundle` attribute.

design/validation-spi.md
vmuzikar: >It doesn't need to be the validators responsibility to decide if sth. is a warning or an error.

+1

What would be the purpose of warning anyway? What's the use case here?

thomasdarimont:
The final validation SPI for profile validations that made it into master is a streamlined version of this with minor adaptations.

Closing this now.

PR #263, opened 2021-04-12
User Profile document updated based on the outcome of the User Profile Working Group

velias:
User Profile design document updated based on the outcome of the User Profile Working Group:
* Mechanism extended to allow more user profiles per realm selected based on client rules
* Added more details about functionality
* Removed some aspects not necessary now (eg converters)


velias:
Updated to reflect changes based on last Thursday's Keycloak team meeting

PR #266, opened 2021-04-21
draft RAR

guymoyo:


tnorimat (commented):
@guymoyo I've read this PR and had not any comment. I agree with your idea of processing authorization details by SPI provider implementation.

design/rich-authorization-requests.md
```diff
@@ -0,0 +1,173 @@
...
+
+* store authorization_details within the grant
+
+````
+rarProcessorProvider.processAuthorizationDetails()
+````
+
+### Admin UI
+
+AS should allow the possibility to select RarProcessor Implementation Module.
```
thomasdarimont, line 158: Should the selection of the RarProcessor implementation / strategy be based only on Realm or also on Client level? 
I can image that different clients might require different details processing strategies, then it would make sense to have be able to override a realm global configuration on client level.

thomasdarimont (commented):
Thanks @guymoyo for your proposal, this looks promising :)

I just have a few remarks inline.

design/rich-authorization-requests.md
```diff
@@ -0,0 +1,173 @@
...
+
+Files/Classes/methods affected:
+
```
thomasdarimont, line 161: I think this should also list the adapter libraries that need to be updated in order to support the new `authorization_details` Authorization request parameter and the `authorization_details` Access token response attribute.

- Keycloak adapter-core (org.keycloak.adapters.OAuthRequestAuthenticator#authenticate)
- Keycloak JS Adapter

design/rich-authorization-requests.md
```diff
@@ -0,0 +1,173 @@
...
+            "iban": "DE02100100109307118603"
+         },
+         "remittanceInformationUnstructured": "Ref Number Merchant"
+      }
+   ]
+````
+
+The type of data contained in authorization details is infinite and depends on the field (Payment, Health, Tax Data, etc...)
+
+Therefore, it will be more scalable to add the processing of authorization details as SPI (e.g. `RarProcessor.Class`). Then, an extension will come with the specific implementation.
```
thomasdarimont, line 46: The spec should also mention that RAR should work with PAR (Pushed Authorization requests for large authentication_details, as described in the [RAR specification for large requests](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar#section-12.4)

guymoyo:
> Thanks @guymoyo for your proposal, this looks promising :)
> 
> I just have a few remarks inline.

Thanks @thomasdarimont, I will check your remarks.

design/rich-authorization-requests.md
```diff
@@ -0,0 +1,173 @@
...
+- Accept `authorization_details` parameter in all places where the `scope` parameter is used
+- Add authorization details to access tokens and token introspection
+
+### Authorization Server Metadata
+
+This parameter must be added to the .well-known/openid-configuration:
+
+- **authorization_details_types_supported**
+
+The authorization data types supported can be determined using the metadata parameter "authorization_details_types_supported", which is an JSON array.
```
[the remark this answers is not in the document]
guymoyo: In my opinion, it's way around, a client should pick the types it wants to use in `authorization_details_types_supported`.  published in AS metadata

ex:

AS `authorization_details_types_supported: https://www.someorg.com/account_information, https://www.someorg.com/payment_initiation`

a client can choose to use just account_information ie `authorization_details_type: https://www.someorg.com/account_information`

Maybe we should also check in the dynamic client registration if `authorization_details_type` are compliant with `authorization_details_types_supported`. because it should normally.

design/rich-authorization-requests.md
```diff
@@ -0,0 +1,173 @@
...
+Therefore, it will be more scalable to add the processing of authorization details as SPI (e.g. `RarProcessor.Class`). Then, an extension will come with the specific implementation.
+
+However, there is some default configuration and behavior which should be implemented:
+
+- Add parameters in Authorization Server Metadata
+- Add parameters in Client Metadata
+- Accept `authorization_details` parameter in all places where the `scope` parameter is used
+- Add authorization details to access tokens and token introspection
+
+### Authorization Server Metadata
```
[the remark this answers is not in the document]
guymoyo: Yes agree, I will add this metadata and set it to `true`.

design/rich-authorization-requests.md
guymoyo: I see. For a more detailed configuration, being able to decide the processing strategies at the client level will be a good thing.
 
I would suggest making the first version at the Realm level and then the other version can be extended with client level.

design/rich-authorization-requests.md
guymoyo: Yes, I will add it.

design/rich-authorization-requests.md
guymoyo: Nice, I completely forgot about the adapters, I will add. However, I do not understand what needs to be changed in [org.keycloak.adapters.OAuthRequestAuthenticator#authenticate](https://github.com/keycloak/keycloak/blob/b1d39aa136662d54deca10c97374f0932ce6316b/adapters/oidc/adapter-core/src/main/java/org/keycloak/adapters/OAuthRequestAuthenticator.java#L263)

design/rich-authorization-requests.md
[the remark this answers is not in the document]
guymoyo: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar#section-10

VinodAnandan:
New revision: https://www.ietf.org/rfcdiff?url2=draft-ietf-oauth-rar-07

design/rich-authorization-requests.md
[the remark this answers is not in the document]
mposolda: +1 to the point that discovery claim `authorization_details_types_supported` will be an aggregated list of all types across all types of processor providers. Regarding collisions, I supposed we can add something like `getPriority` to the processor methods? This will allow that custom processor provider will be chosen for processing when dealing with the particular conflicting type.

For example: Assume that there will be built-in provider `DefaultProcessor`, which will support types `my-type-1` and `my-type-2` with priority 1. Then assume that administrator will deploy his own custom provider with higher priority (EG. `MyCustomProcessor`), which will support types `my-type-2` and `my-type-3`  and will have higher priority.

This will ensure that:
- The type `my-type-1` will be processed by default processor
- The type `my-type-3` will be processed by MyCustomProcessor
- The type `my-type-2` exists in both processors, but since `MyCustomProcessor` has higher priority, it will be chosen for processing this request.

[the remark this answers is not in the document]
mposolda: It seems to me that priority of the processor can be either hardcoded in the implementation or overriden in the standalone(-*).xml . Which means configurable for the whole server. It seems to me that have the processor priority specific for the whole server is fine for most of the use-cases and we don't need the realm or client level.

@thomasdarimont @guymoyo WDYT? Do you have any use-case for have the ability to choose the processor for the realm or client level?

design/rich-authorization-requests.md
mposolda: This is the question... I was thinking if it is sufficient to stick with the server level for choosing the processing strategy?

Or do you have some use-cases in mind when different clients (or different realms) may want to stick with the different processing requirements?

IMO the `authorization_details` is more advanced version of the `scope` parameter with more fine-grained support. And for example for `scope` parameter, I did not see use-case when different clients want same scope with different meaning. For example that client1 in theory may want to use `scope=profile` to have a bit different meaning than client2 using `scope=profile` . In theory, client1 may want a bit different data to be shown for the `profile` than client2. However in reality, this was never needed...

ATM it seems to me that for choosing the processor, we may also stick to the server level (or provider level) as I pointed in the other place and add more fine-grained control if there is later the use-case for it?

mposolda: One thing, which seems to me will be good for admin UI, might be to being able to choose supported types of authorization_detail types for various clients? This is supported in the client registration parameter `authorization_details_types` and hence probably also need some control in the admin console.

I can imagine also some support for evaluating access tokens (The stuff when you go to client -> Tab `Client Scopes` -> click to some client scope -> tab `Evaluate` to generate example tokens. It can be interesting to be able to generate tokens for various provided `authorization_details` ). However this support for evaluating example token is something to be done as a follow-up PR if needed (Not necessarily in the first PR of this feature IMO).

design/rich-authorization-requests.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Documentation
+RAR usage should be properly documented.
+
+Affected documents: Securing Applications and Services Guide
+
+## Resources
+* [draft-ietf-oauth-rar][1]
+
+[1]: https://tools.ietf.org/html/draft-ietf-oauth-rar-04
```
mposolda, line 196: FYI. Just a minor note that newest draft is https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar-07 ...

mposolda (changes requested):
Finally added review to this specification. Sorry for being late...

I've added few points inline and marking this as "Request changes" until these are discussed and addressed somehow.

One general note: There is other specification FAPI Lodging Intent https://bitbucket.org/openid/fapi/src/master/Financial_API_Lodging_Intent.md . It seems to me that specification tries to address very similar problem like this RAR specification (Insufficient capabilities of the OAuth2 `scope` parameter and the ways how to overcome those limited capabilities). Will be probably good to also take the other specification into account and makes sure that they are somehow compatible with each other. The specifications are both drafts and both have same/similar authors, so it is probably not two completely independent efforts :-) 

It seems to me that some intersection for these specifications is for example generating the consent screen (as it seems that RAR may also need some capabilities on the consent screen).

tnorimat:
@mposolda 
For https://github.com/keycloak/keycloak-community/pull/266#pullrequestreview-757411929 , [OIDF's FAPI](https://openid.net/wg/fapi/) site mentions that FAPI 1.0 lodged intent be replaced with OAuth PAR + OAuth RAR. Therefore, at this time, it might be OK for us to focus only on supporting RAR.

tnorimat:
@guymoyo , @mposolda , @thomasdarimont ,
As an design document, it seems to be OK to be merged after the current review comments are resolved.

IMO, when implementing it, the following points might be difficult to handle.
 - showing an consent screen with respect to "authroization_details"
 - enriching "authroization_details" on UI
 - managing consents with respect to "authroization_details"
 
The last point is also related to [dynamic scopes](https://groups.google.com/g/keycloak-dev/c/yGCWth8reNw/m/TEDYZS8hAwAJ) and [FAPI 2.0 Grant Management for OAuth 2.0](https://openid.net/specs/fapi-grant-management-01.html).

I think the current keycloak's consent model may need to be modified.

Also, on implementation, I think keycloak need not include any concrete RAR processor provider because it is highly dependent on specific use-cases. However, arquillian integration tests include some concrete RAR processor.

design/rich-authorization-requests.md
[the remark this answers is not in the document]
guymoyo: Hi @mposolda, thanks for the explanation, it's clear to me. I don't have a particular use case at the realm or client level.
I will add your text as an explanation for `authorization_details_types_supported`.

design/rich-authorization-requests.md
guymoyo: Hi @mposolda , as mentioned in the previous comment processor will be selected at the server level. And we will add the possibility in the admin UI to select the authorization_detail types for a client.

design/rich-authorization-requests.md
guymoyo: yes, I looked at the difference, it's more some text. The concept didn't change and no new parameter.

mposolda:
> @guymoyo , @mposolda , @thomasdarimont , As an design document, it seems to be OK to be merged after the current review comments are resolved.
> 
> IMO, when implementing it, the following points might be difficult to handle.
> 
>     * showing an consent screen with respect to "authroization_details"
> 
>     * enriching "authroization_details" on UI
> 
>     * managing consents with respect to "authroization_details"
> 
> 
> The last point is also related to [dynamic scopes](https://groups.google.com/g/keycloak-dev/c/yGCWth8reNw/m/TEDYZS8hAwAJ) and [FAPI 2.0 Grant Management for OAuth 2.0](https://openid.net/specs/fapi-grant-management-01.html).
> 
> I think the current keycloak's consent model may need to be modified.
> 
> Also, on implementation, I think keycloak need not include any concrete RAR processor provider because it is highly dependent on specific use-cases. However, arquillian integration tests need to include some concrete RAR processor.

@tnorimat These are very good points. It will be good to consider RAR as a more powerful alternative to classic "scope" parameter, however dynamic scopes have lots of common things with RAR.

It turns that Keycloak team has some internal requirements and it looks that it may need to prioritize the work on dynamic scopes soon. There are some use-cases to dynamic scopes discussed here https://github.com/keycloak/keycloak/discussions/8486 and some other discussion started here https://github.com/keycloak/keycloak/discussions/8488#discussioncomment-1394572 .

Overally, we will need to figure some way how to do RAR and dynamic scopes together and make sure that whatever is done will match the requirements for both these use-cases.

For now, I am not 100% sure how to move forward. I suggest to anyone to comment on the mentioned discussions for dynamic scopes and eventually add feedback to them.

stianst:
I haven't had time to read through this proposal yet, but I've come to the conclusion that we must not extend "scopes" without considering RAR, and at the same time we must not implement RAR as a separate thing. The two have very similar needs: a) client includes some additional details in authz request, b) some policies decide if it's permitted or not, c) the user optionally consents to some stuff, and d) some claims are added to tokens.

PR #267, opened 2021-04-21
added basis for MTLS Extension design 

DmitryMishchuk:
added basis for MTLS Extension design  for additional introspection and validation of SSL Certificates

### DIFF
--- a/design/mtls-validation-extension.md
+++ b/design/mtls-validation-extension.md
@@ -1,0 +1,46 @@
+# MTLS Validation Extension (MTLS_Ext)
+
+* **Status**: Notes
+* **JIRA**: TBD
+
+
+## Motivation
+
+MTLS validation extension is designed to enable extension of standard SSL Certificate validation already present inside Keycloak (X509) enabling validation of additional fields present in different certificate types (example: QWAC), when MTLS 
+support is
+enabled.
+
+QWAC is a sample of standard certificate extension that is occurring on OPEN BANKING Environment 
+(link to [QWAC spec](https://www.etsi.org/deliver/etsi_ts/119400_119499/119495/01.03.01_60/ts_119495v010301p.pdf)).
+
+Introducing an extra introspection/validation SPI has such main benefits:
+
+* Support validation of additional SSL Certificate fields
+* Support for market specific OpenBanking implementations for example European PSD2 (QWAC) or Australian CDR (normative usage of existing fields)
+* Support custom (proprietary) certificate validation implementations
+
+## Implementation details
+We have to provide new Certificate validation SPI.
+This SPI may be implemented by a separate service (plugin), which deals with custom SSL Certificate processing.
+This plugin should be deployed inside the Keycloak and should handle all custom (proprietary) validation and parsing logic, so that Keycloak only interacts with it via SPI.
+
+### Affected classes
+- X509ClientCertificateAuthenticator (existing class bearing the basic SSL validation and storage logics)
+- ExtendedCertificateValidation (SPI meant to build validation plugins for different SSL Certificate types)
+
+### X509ClientCertificateAuthenticator and UI implementation part
+- Admin UI section `Client/Advanced Settings/OAuth 2.0 Mutual TLS Certificate Bound Access Tokens Enabled` should be extended with a flag `Enable custom certificate validation` to be sure that additional validation should be turned on/off for
+  particular client
+- Adding ExtendedCertificateValidation.class as SPI for plugin extension
+- Modification of X509ClientCertificateAuthenticator to with calls to ExtendedCertificateValidation SPI to introspect/validate the certificate if `custom validation is enabled` for current `client`
+
+### ExtendedCertificateValidation SPI class
+Sample implementation for QWAC SSL Certificate validation must include:
+Method `introspect` must handle extraction and storage of additional fields from custom certificates into the AuthenticationFlowContext
+Method `validate` must handle all the logics concerning certificate validation
+
+Additionally, it may carry the following logics:
+- additional fields validation
+- roles validation for presence and syntax
+- correspondence of request(PAR/RAR) to effective roles in certificate
+- online check for certificate validity at certificate issuer


DmitryMishchuk:
@tnorimat Please review

design/mtls-validation-extension.md
```diff
@@ -0,0 +1,46 @@
...
+Introducing an extra introspection/validation SPI has such main benefits:
+
+* Support validation of additional SSL Certificate fields
+* Support for market specific OpenBanking implementations for example European PSD2 (QWAC) or Australian CDR (normative usage of existing fields)
+* Support custom (proprietary) certificate validation implementations
+
+## Implementation details
+We have to provide new Certificate validation SPI.
+This SPI may be implemented by a separate service (plugin), which deals with custom SSL Certificate processing.
+This plugin should be deployed inside the Keycloak and should handle all custom (proprietary) validation and parsing logic, so that Keycloak only interacts with it via SPI.
```
tnorimat, line 25: I think that the ExtendedCertificateValidation provider is applied to the client. How do you apply this provider onto the client?
- per keycloak
   One type of ExtendedCertificateValidation provider is applied to all client in all realm in the keycloak.
- per realm
   One type of ExtendedCertificateValidation provider are applied to all client in the realm (eg. realm A) and other one is applied to all client in the other realm (eg. realm B)
- per realm and client
  One type of ExtendedCertificateValidation provider is applied to some client (eg. Client A) in the realm (eg. realm A) and other one is applied to other client (Client B) in the this realm.

I think `per keycloak` is the easiest to implement and it is reasonable because it is not so plausible that keycloak treats several market specific profiles (PSD2, UK OB, CDR) simultaneously (eg. realm A is for PSD2, realm B is for UK OB, realm C is for CDR ...).

design/mtls-validation-extension.md
```diff
@@ -0,0 +1,46 @@
...
+- X509ClientCertificateAuthenticator (existing class bearing the basic SSL validation and storage logics)
+- ExtendedCertificateValidation (SPI meant to build validation plugins for different SSL Certificate types)
+
+### X509ClientCertificateAuthenticator and UI implementation part
+- Admin UI section `Client/Advanced Settings/OAuth 2.0 Mutual TLS Certificate Bound Access Tokens Enabled` should be extended with a flag `Enable custom certificate validation` to be sure that additional validation should be turned on/off for
+  particular client
+- Adding ExtendedCertificateValidation.class as SPI for plugin extension
+- Modification of X509ClientCertificateAuthenticator to with calls to ExtendedCertificateValidation SPI to introspect/validate the certificate if `custom validation is enabled` for current `client`
+
+### ExtendedCertificateValidation SPI class
```
tnorimat, line 37: We need to set up ExtendedCertificateValidation provider even if this provider does not have any configuration items. To do so, I think we write such configuration on the setting files (eg. standalone.xml) or executing jboss cli's command. Is that right?

tnorimat (commented):
@DmitryMishchuk I've reviewed this document and added some comments. Could you check them?

design/mtls-validation-extension.md
```diff
@@ -0,0 +1,46 @@
...
+We have to provide new Certificate validation SPI.
+This SPI may be implemented by a separate service (plugin), which deals with custom SSL Certificate processing.
+This plugin should be deployed inside the Keycloak and should handle all custom (proprietary) validation and parsing logic, so that Keycloak only interacts with it via SPI.
+
+### Affected classes
+- X509ClientCertificateAuthenticator (existing class bearing the basic SSL validation and storage logics)
+- ExtendedCertificateValidation (SPI meant to build validation plugins for different SSL Certificate types)
+
+### X509ClientCertificateAuthenticator and UI implementation part
+- Admin UI section `Client/Advanced Settings/OAuth 2.0 Mutual TLS Certificate Bound Access Tokens Enabled` should be extended with a flag `Enable custom certificate validation` to be sure that additional validation should be turned on/off for
```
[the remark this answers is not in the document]
DmitryMishchuk: Is the feature `OAuth 2.0 Mutual TLS Certificate Bound Access Tokens Enabled` irrelevant so we should remove it from UI?
I actually thought these features are bound to each other with mTLS usage, so idea was extending/adding to it. Please correct me if I'm wrong.
The pain idea for the feature was to parse/validate certificate as PoP, not issuing tokens from certificates but adding another level of protection to tokens, though usage of token would be impossible without corresponding certificate bound to it.

design/mtls-validation-extension.md
DmitryMishchuk: I actually thought of third option, as you surely might have situations when you need same realm work different for different client example: 1 client works with some backend through back channel using some grant password flow without anything other protection means, 2 client works for some XS2A like open interface with FAPI and mTLS enabled, 3 client is for OnlineBanking OAuth2 browser flow without mTLS support as there can't be any certificate at your own customer disposal.
And yes, you're absolutely right that implementing global support is the easiest option. Here I did not mention possibility of having more than one `extended validator`, but there could be people willing to have multiple additional validators and picking the desired one from UI for respective client

design/mtls-validation-extension.md
DmitryMishchuk: Yes, thank you. Adding SPI implies adding `providers` and some configurations for such.

design/mtls-validation-extension.md
tnorimat: Yes, I agree with your proposal, namely the option 3.

design/mtls-validation-extension.md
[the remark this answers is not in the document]
tnorimat: No, we need not to remove `OAuth 2.0 Mutual TLS Certificate Bound Access Tokens Enabled` on `Client/Advanced Settings`. I suggested that `Enable custom certificate validation` flag would be added to `Client/Advanced Settings`.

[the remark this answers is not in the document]
DmitryMishchuk: Yes that would be the correct place for it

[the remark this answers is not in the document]
DmitryMishchuk: Ah, now I see what you meant. Done.

PR #305, opened 2021-08-27
Fix device authorization grant endpoints

matgnt:
Fix the endpoint URLs according to current version (15.0.0) to make it easier to get started with the examples

### DIFF
--- a/design/oauth2-device-authorization-grant.md
+++ b/design/oauth2-device-authorization-grant.md
@@ -12,7 +12,7 @@
 
 ### Device Authorization Endpoint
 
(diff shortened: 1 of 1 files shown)


mposolda (approved):
@matgnt Thanks for the correction! FYI. Device authorization grant is already implemented, but I suppose you are aware of it.

matgnt:
Thanks for merging. Yes, I already tried it. Since it is not documented very well, I came across the doucmentation here and it took me a while to find out the endpoints were not correct :-) It should be easier for others now to go through the examples here.
Thanks,
Matthias

PR #318, opened 2021-10-01
Namespaced Roles

dgozalo:
Creating a separate PR as this proposal is too large for having the whole text in Github discussions as defined in https://github.com/keycloak/keycloak/pull/8503/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R48

Coming from: https://github.com/keycloak/keycloak/discussions/8516

### DIFF
--- a/design/namespaced-roles.md
+++ b/design/namespaced-roles.md
@@ -1,0 +1,299 @@
+* **Status**: Draft #1
+* **JIRA**: TBD
+* **Github Discussion**: [Namespaced Roles](https://github.com/keycloak/keycloak/discussions/8516)
+
+## Motivation
+
+In Keycloak, roles are defined at the Realm and Client levels. These roles can then be assigned directly to a User or to a Group, making every user in that group inherit the role.
+
+This is done through role mapping tables in the database, limiting the way you can assign roles to those two use cases.
+
+We will need more flexibility in the future, for example, to assign a role to a user in the context of a specific group (instead of having everyone in the group inherit a role) or, as the tenancy model gets implemented, a way to both define and assign roles in the context of a tenant.
+
+In order to have this flexibility and still support current use cases, we could create a new kind of role or backport existing roles to this new format: Namespaced roles.
+
+## Specification
+
+A Namespaced Role would follow a certain structure that would unambiguously identify it across the system, following a hierarchical structure.
+
+They would work similarly to ARNs in AWS, where a portion of the format is fixed and always identifiable by position, with the possibility of containing empty fields, containing a final variable part that could be freely named.
+
+An example would be:
+
+`role_v1:/\<realm\>/\<tenant\>/\<client\>/\<group\>/<roleName>`
+
+**Realm**: All resources in Keycloak are namespaced to a realm, so a unique identifier for a role would also contain it.
+   - It doesn't support empty values - all roles will be scoped to a realm.  
+   - Example: `role_v1:/realm1////roleName`
+
+**Tenant**: There's a proposal to make Keycloak multitenant aware, so even if it's not currently supported, for future compatibility, it should be considered when defining a unique identifier.
+   - Supports empty values - A role might be created and assigned at a realm level
+   - Example: `role_v1:/realm1/tenantA///roleName`, `role_v1:/realm1///groupA/roleName`
+
+**Client**: Roles can be defined and assigned for a certain client in Keycloak, so a unique identifier has to support this possibility.
+   - Supports empty values - A role might be created and assigned without being restricted to a client.
+   - Example: `role_v1:/realm1/tenantA/clientId//roleName`, `role_v1:/realm1/tenantA//groupA/roleName`
+
+**Group**: Roles can be assigned to groups or, as intended in this design proposal, assigned to a user in the context of a group.
+   - Supports empty values - A role might be created at any level and not necessarily assigned to a group.
+   - Example: `role_v1:/realm1/tenantA/clientId/groupB/roleName`, `role_v1:/realm1/tenantA/clientId//roleName`
+
+**RoleName**: Finally, the name of the role that is being defined.
+   - Doesn't support empty values - You need to know the name of the role that is being created.
+   - Example: `role_v1:/realm1/tenantA/clientId/groupB/roleName`
+
+A possible free-form role identifier could be allowed for resource-server specific roles.
+
+The initial fixed part would be kept in order to identify the role within a realm, tenant, client, group etc, but instead of a fixed name for the role, it would be a reference to other hierarchical role formed with any random structure.
+
+A possible format would be:
+
+- `role_v1:/realm1/tenant1//group1/frole_v1:/mycompany/resources/department-a-roles/developer`
+
+If the role name is another reference to a free form role, starting with `frole_v1:/`, this role will also be represented in a hierarchical way, making it possible for protocol mappers to show them as such in tokens.
+
+## Use Cases
+
+As previously mentioned, there are a few use cases that we'll want to support and are currently not possible in Keycloak.
+
+### Use Case 1 - Assign a role in the context of a group
+
+Say we want to have the following group composition in our realm (or our tenant, in the future), most likely imported from a federated LDAP:
+
+realm/tenant:tenant1
+ - group:iam
+ - group:devops
+ - group:leadership
+
+Following the current implementation, the users would need to define roles at the realm level and then assign them to these groups. Everyone in those groups would inherit the roles.
+
+But for a client, these groups may have business meaning or reflect organizational structure within their company, they aren't just a mean to group users together with the same roles, but rather fully functional teams that users can be part of with different roles.
+
+So UserA would be a `manager` in `group:iam`, but UserB could be a `developer`in `group:iam`.
+
+Namespaced roles would be specified when created, and loosely assigned based on the namespace, so let's say we have the following roles:
+
+`role_v1:/realm1/tenant1//iam/manager`
+`role_v1:/realm1/tenant1//iam/developer`
+`role_v1:/realm1/tenant1//devops/developer`
+`role_v1:/realm1/tenant1//devops/devops_role`
+
+Now these roles are assigned directly to `UserA` and `UserB` so that:
+
+- `UserA` has the `role_v1:/realm1/tenant1//iam/manager` and the `role_v1:/realm1/tenant1//devops/developer` roles.
+- `UserB` has the `role_v1:/realm1/tenant1//iam/developer` and the `role_v1:/realm1/tenant1//devops/developer` roles.
+
+Also, we should be able to still support giving everyone in a group an inherited role, such as:
+
+- `Group:devops` has the `role_v1:/realm1/tenant1//devops/devops_role` role.
+
+#### Representation in the tokens
+
+So now that we have a namespaced role attached to a user, the representation becomes inherently hierarchical, a **possible representation** in the tokens could look like:
+
+- For UserA
+
+```json
+{
+    ....
+    "roles": {
+        "tenants": {
+            "tenant1": {
+                "groups": {
+                    "iam": ["manager"],
+                    "devops": ["developer", "devops_role"]
+                }
+            }
+        }
+    }
+    ....
+}
+
+```
+
+- For UserB
+
+```json
+{
+    ....
+    "roles": {
+        "tenants": {
+            "tenant1": {
+                "groups": {
+                    "iam": ["developer"],
+                    "devops": ["developer", "devops_role"]
+                }
+            }
+        }
+    }
+    ....
+}
+```
+
+This also provides a way to isolate certain roles to specific tenants.
+
+- `role_v1:/realm1/tenant1//iam/developer` would only be visible in `tenant1` while
+- `role_v1:/realm1/tenant2//iam/somethingelse` would only be visible in `tenant2`.
+
+Or just to specific groups within the realm:
+
+- `role_v1:/realm1///iam/developer` would only be assignable to users in Group `iam`.
+- `role_v1:/realm1///devops/devops` would only be assignable to users in Group `devops`.
+
+### Use Case 2 - Custom, resource server specific roles
+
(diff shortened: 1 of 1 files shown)


design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+* **JIRA**: TBD
+* **Github Discussion**: [Namespaced Roles](https://github.com/keycloak/keycloak/discussions/8516)
+
+## Motivation
+
+In Keycloak, roles are defined at the Realm and Client levels. These roles can then be assigned directly to a User or to a Group, making every user in that group inherit the role.
+
+This is done through role mapping tables in the database, limiting the way you can assign roles to those two use cases.
+
+We will need more flexibility in the future, for example, to assign a role to a user in the context of a specific group (instead of having everyone in the group inherit a role) or, as the tenancy model gets implemented, a way to both define and assign roles in the context of a tenant.
```
stianst, line 11: Would also like the motivation to touch on things like:

* The need to create clients just to serve as a placeholder for roles, like we do for realm management roles
* The inability to share a common set of roles with a group of clients
* The inability to properly organize roles

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Specification
+
+A Namespaced Role would follow a certain structure that would unambiguously identify it across the system, following a hierarchical structure.
+
+They would work similarly to ARNs in AWS, where a portion of the format is fixed and always identifiable by position, with the possibility of containing empty fields, containing a final variable part that could be freely named.
+
+An example would be:
+
+`role_v1:/\<realm\>/\<tenant\>/\<client\>/\<group\>/<roleName>`
```
stianst, line 23: This is a bit hard to read due to what I assume is escaping.
```suggestion
`role_v1:/{realm}/{tenant}/{client}/{group}/{roleName}
```

Roles are already scoped to a realm, so that shouldn't be included in the role namespace.

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+
+**Realm**: All resources in Keycloak are namespaced to a realm, so a unique identifier for a role would also contain it.
```
stianst, line 25: Roles are already tied to a realm, so realm shouldn't be part of the namespace I'd say.

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+
+Following the current implementation, the users would need to define roles at the realm level and then assign them to these groups. Everyone in those groups would inherit the roles.
+
+But for a client, these groups may have business meaning or reflect organizational structure within their company, they aren't just a mean to group users together with the same roles, but rather fully functional teams that users can be part of with different roles.
+
+So UserA would be a `manager` in `group:iam`, but UserB could be a `developer`in `group:iam`.
+
+Namespaced roles would be specified when created, and loosely assigned based on the namespace, so let's say we have the following roles:
+
+`role_v1:/realm1/tenant1//iam/manager`
```
stianst, line 76: I don't understand the meaning of the double forward slash (//). I also somewhat think that this is being complicated by including the tenant in the namespace. Perhaps a simpler example for role within a group is just "role_v1:/iam/manager'

design/namespaced-roles.md
stianst, line 23: In my mind a role namespace was following the syntax:

[namespace/]role-name

```
kc/clients/myclient/myrole
```

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+
+As previously mentioned, a role would be namespaced to a certain client such as:
+
+-  `role_v1:/realm1//clientId//roleName`
+
+A different way to be more flexible on giving users the possibility to set their own resource-server specific roles would be to provide a free-form unique role identifier that can be parsed by protocol mappers or just shown in the tokens in a hierarchical way.
+
+For example, a free form role could look like:
+
+- `role_v1:/realm1/tenant1///frole_v1:/mycompany/resources/department-a-roles/developer`
```
stianst, line 157: I'm struggling a bit to understand this syntax. What does '////' imply, and why is there 'frole_v1:/'

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+
+The initial fixed part would be kept in order to identify the role within a realm, tenant, client, group etc, but instead of a fixed name for the role, it would be a reference to other hierarchical role formed with any random structure.
+
+A possible format would be:
+
+- `role_v1:/realm1/tenant1//group1/frole_v1:/mycompany/resources/department-a-roles/developer`
+
+If the role name is another reference to a free form role, starting with `frole_v1:/`, this role will also be represented in a hierarchical way, making it possible for protocol mappers to show them as such in tokens.
+
+## Use Cases
```
stianst, line 55: Would be good to explicitly mention how role namespaces can clean-up the roles used "internally" by Keycloak's admin APIs. Getting rid of the strange clients, and having a separate/dedicated namespace for the roles so they don't conflict with "custom roles". I had something like the following in mind:
```
kc/admin/{realm-name}/{role-name}
```

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+
+- `role_v1:/realm1/tenant1//iam/developer` would only be visible in `tenant1` while
+- `role_v1:/realm1/tenant2//iam/somethingelse` would only be visible in `tenant2`.
+
+Or just to specific groups within the realm:
+
+- `role_v1:/realm1///iam/developer` would only be assignable to users in Group `iam`.
+- `role_v1:/realm1///devops/devops` would only be assignable to users in Group `devops`.
+
+### Use Case 2 - Custom, resource server specific roles
```
stianst, line 143: Would be good to touch on how role namespaces align with client roles. In my mind client roles should be migrated to role namespaces, where a dedicated role namespace can easily be created for a specific client, or for a group of clients. With also making it more clear that you don't need to create a client just to represent some roles for a resource server.

design/namespaced-roles.md
```diff
@@ -0,0 +1,299 @@
...
+Now these roles are assigned directly to `UserA` and `UserB` so that:
+
+- `UserA` has the `role_v1:/realm1/tenant1//iam/manager` and the `role_v1:/realm1/tenant1//devops/developer` roles.
+- `UserB` has the `role_v1:/realm1/tenant1//iam/developer` and the `role_v1:/realm1/tenant1//devops/developer` roles.
+
+Also, we should be able to still support giving everyone in a group an inherited role, such as:
+
+- `Group:devops` has the `role_v1:/realm1/tenant1//devops/devops_role` role.
+
+#### Representation in the tokens
```
stianst, line 90: Would like to see some alternatives on how roles can be represented in a token. For example options to flatten them if that makes sense. For example a client may not understand the concept of role namespaces, but would only want the roles for a specific namespace flattened.

Also, would be good to touch on how the roles can be represented in the current way they are being done in tokens with regards to realm and client roles.

design/namespaced-roles.md
dgozalo: The double slashes are basically a missing field.

Take `role_v1:/tenant1/client1/iam/manager`, that contains values for all fields, because it's created in the scope of a client.

Now, imagine you wanted to create the same role, but without the client, in order to keep the order of variables and make it possible to retrieve by position, you need to keep an empty space between tenant and group, hence:

`role_v1:/tenant//iam/manager`.

design/namespaced-roles.md
dgozalo: This follows the same format, but the namespace is composed of a fully identifiable path, made up by the scopes where the role is created.

This is to make it possible to decompose a role by its namespace and extract variables from it and to make querying them easy.

design/namespaced-roles.md
dgozalo: I'll add some more explanation on what the the `frole` is. But it's essentially a custom namespace created by users, while keeping the fixed path that contains all the variables to identify a role.

The forward slashes are basically there to keep separation and make it possible to access a variable by its position in the path.

design/namespaced-roles.md
dgozalo: Good point. I think the biggest struggle in this whole namespaces spec is how to fit clients in.

We want to create a namespace that can be seen as a unique path to a role, but clients are not really part of that hierarchy necessarily.

design/namespaced-roles.md
dgozalo: +1. I'll add those points.

design/namespaced-roles.md
dgozalo: A much better representation indeed. I'll change it.

design/namespaced-roles.md
dgozalo: Good point, it's also even a field in the database, will change it.

design/namespaced-roles.md
stianst: What I had in mind wasn't to have some special meaning of the namespace, but rather it's just a free-form hierarchy. So you would just use `role_v1:/tenant/iam/manager` or `role_v1:/iam/manager`. Up to you. I was thinking we'd have a controlled namespace for internal roles like `role_v1:/kc/clients/{client-name}/{role-name}`

design/namespaced-roles.md
stianst: Yeah, I think I agree with having the role_name first is better.

design/namespaced-roles.md
stianst: I don't think we need/should have the fixed path part at all to be honest.

design/namespaced-roles.md
stianst: What I had in mind is that we'd reserve 'kc' as a special internal namespace, and for a dedicated/automatic namespace for a client it would just be `{role}:/kc/clients/{client-name}`

design/namespaced-roles.md
```diff
@@ -0,0 +1,349 @@
...
+
+Some of the Keycloak supported entities would be:
+
+**Tenant**: There's a proposal to make Keycloak multitenant aware, so even if it's not currently supported, for future compatibility, it should be considered when defining a unique identifier.
+   - Example: `role_v1:/kc/tenants/tenantA/roleName`.
+
+**Client**: Roles can be defined and assigned for a certain client in Keycloak, so a unique identifier has to support this possibility.
+   - Example: `role_v1:/kc/clients/clientId/roleName`.
+
+Alternatively, if a namespace wants to be created for roles that can be shared among clients, it could be defined such as:  `role_v1:/clients/*/roleName`.
```
stianst, line 51: Not sure I understand why there's a `*` here

stianst: Wonder if in this case you'd just use a free-form role namespace instead.

design/namespaced-roles.md
```diff
@@ -0,0 +1,349 @@
...
+
+A free-form role identifier would be allowed for resource-server specific roles thay may hold business meaning for users. It would lack the fixed `kc` or `ud` particles and allow any user defined path. For example:
+
+- `role_v1:/business-units/myBusinessUnit/const-centers/myCostCenter/roleName`
+
+**Question**: In the initial proposal, these roles would be preceded by a fixed part that would identify the context where this role would be applied. We could do somethng similar here such as:
+
+- `role_v1:/groups/{group-name}/role-path/business-units/myBusinessUnit/const-centers/myCostCenter/roleName`
+
+But not convinced on the way to let Keycloak know that that's not a role name but rather a new namespace.
```
stianst, line 66: I was thinking that clients, groups, tenants, etc, would instead of having a fixed path to their namespace would allow configuring the namespace that is associated with them.

So, a client could have it's default namespace be `/kc/clients/{client-id}`, but could be configured to use a different namespace like `/my-roles/my-group-of-apps`

design/namespaced-roles.md
```diff
@@ -0,0 +1,349 @@
...
+
+These roles would only exist in the Master realm and would have a meaning for Keycloak, hence the `kc` fragment.
+
+- `role_v1:/kc/admin/{role-name}` -> Roles to manage the Keycloak instance.
+
+When a new Realm is created in Keycloak, a new client is created in the Master realm to represent a namespace for this new Realm's admin roles. With the exception of the previous one, they will be used to deprecate the Clients that Keycloak creates to group certain management roles together.
+
+The rest of these roles would always be scoped to a specific realm that would replace the current client-per-realm.
+
+- `role_v1:/kc/realms/{realm-name}/clients/{client-name}/{role-name}` -> Roles to manage a certain client.
```
stianst, line 78: This is more fine-grained than what we have today. Right now, we have just manage-clients, not ability to manage an individual client. Wonder if this level is too fine-grained for roles, and rather we should look at fine-grained admin permission to manage individual clients (and also could look at integrating with external authz services in the future).

design/namespaced-roles.md
```diff
@@ -0,0 +1,349 @@
...
+
+Right now in Keycloak, clients are effectively a representation of a Resource Server (among other things). You can assign roles to clients and those roles would only be usable on those clients.
+
+As previously mentioned, a role would be namespaced to a certain client such as:
+
+-  `role_v1:/ud/clients/{client-name}/{role-name}`
+
+or with a regexp, they could be namespaced to a specific subset of clients:
+
+- `role_v1:/ud/clients/{regexp-containing-wildcard}/{role-name}`
```
stianst, line 216: Wouldn't this be incredibly complex to resolve? Also, we'd want to use client-id to specify a client rather than client-name, and often these are just generated (and really should be), so is there a meaningful regex here?

design/namespaced-roles.md
```diff
@@ -0,0 +1,349 @@
...
+                "department-a-roles": ["developer"]
+            }
+        }
+    }
+    ....
+}
+```
+Or even scope this custom namespace to a specific client or client sets:
+
+- `role_v1:/ud/clients/{client-name-or-regexp}/frole:/mycompany/resources/department-a-roles/developer`
```
stianst, line 241: Hmmm... I'm struggling to understand this one ;)

PR #328, opened 2021-10-19
[KEYCLOAK-19303] - Allow managing providers through CLI

pedroigor:


design/keycloak.x/configuration.md
```diff
@@ -563,6 +563,89 @@ Examples of commands are:
...
+
+To get more information, including the complete list of providers for a SPI, append `--full` to your command line.
+
+To get information about providers for a specific SPI, append `--spi <spi>` to your command line.
+```
+
+To get more information about each SPI, it should be possible to run `kc.[sh|bat] list-spi`: 
+
+```
+Listing available SPI (see --help for more options)
```
DGuhr, line 599: I'm not sure if this makes sense at all without the information provided by --full. Can you clarify the use case of just running `kc.sh list-spi`?

stianst (changes requested):
As an extension often comes in the form of implementing more than a single provider I would like to introduce a concept that can wrap multiple providers into an extension, and have the option to manage extensions, rather than just individual providers.

PR #330, opened 2021-12-15
Keycloak.X Operator

vmuzikar:


design/keycloak.x/operator.md
```diff
@@ -0,0 +1,121 @@
...
+We will be leveraging Kubernetes volumes to act as "caches" for the [augmented/configured](https://quarkus.io/guides/reaugmentation) version of Keycloak.
+An initial POC to show the concept has been drafted here:
+https://github.com/andreaTP/poc-mutable-jar
+
+We will use Kubernetes volumes to cache the augmented version of the binaries.
+The artifacts in the kubernetes volume will be produced by an init-container and the operation is going to result in a noop in case the volume has already been populated by a compatible augmentation.
+
+### Connecting to a database
+
+A Postgres DB will have to be provisioned externally, it's not Keycloak Operator's responsibility to manage a database. The DB credentials will be stored as K8s Secrets.
```
kfox1111, line 84: Some more detail here would be useful. there are usually 2 things called databases. The software instance, and the stuff inside (schema, tables, etc). I'm guessing your talking about not managing the software instance but managing the rest. For upgrades, the stuff in the db may need things done to it in particular orders to get a proper upgrade. that will be supported?

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,121 @@
...
+
+In long-term plan we'll add a limited integration with a Postgres Operator to leverage its backup functionalities for Keycloak upgrades.
+
+### Observability
+
+The operator will provide CR metrics as well as it will provide integration with Prometheus, Grafana and AlertManager for both operator and operand metrics. This will be addressed in an upcoming design proposal.
+
+### Ingresses
+
+The operator will provide an out-of-the-box experience using an opinionated default Ingress configuration.
```
kfox1111, line 94: That can be disabled if desired?

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,121 @@
...
+* Vanilla Kubernetes >=1.20
+
+Other Kubernetes distributions are supported only in the best effort mode.
+
+## Distribution
+
+The Operator deployment is going to be performed leveraging OLM providing both CLI approach via `Subscription` objects for managing the operator installation, and UI in OpenShift. The Operator as such is going to be distributed as a container image.
+
+## Migration from the old operator
+No direct migration path. Generic migration steps will be documented.
```
kfox1111, line 115: Thats ok. But is there a guarantee both can co'exist in the same cluster?

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Static configuration through Git
+
+The `Keycloak` CRD will enable defining a specific commit (identified by an hash) in a Git repository containing the static configuration in the form of JSON/YAML files. To update the configuration the user will simply change a commit hash in a `Keycloak` CR and the operator will roll out the new configuration to all pods.
+
+#### Static configuration through Kubernetes API
+
+The operator will leverage dedicated CRD(s), initially, there will be only one `Realm` CRD directly translated from Keycloak's [RealmRepresentation](https://github.com/keycloak/keycloak/blob/c7134fd5390d7c650b3dfd4bd2a2855157042271/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java). A Realm includes all subresources. As a result, it is going to be possible to configure every object in Keycloak through this CR even though for some of them it won't be recommended (e.g. Users). To implement this, the operator will simply translate the CRs to YAML files and mount them to Keycloak pods, again leveraging Store.X.
+
```
kfox1111, line 57: Will there be parallel api support so the both operators can be used in parallel?

Please ensure secrets never are stored in the CR. This is a problem with the current api. This should not be repeated in the new api.

kfox1111 (changes requested):
Looking pretty good. some comments inline.

design/keycloak.x/operator.md
hecsalazar: one ingress that would be nice to have is an Istio Gateway + VirtualService as entry point

design/keycloak.x/operator.md
vmuzikar: The two (old written in Go, new written in Java) won't be (most probably) possible to use at the same time. There'd be API conflicts (CRDs) and it wouldn't make much sense as the old one supports only Wildfly dist of Keycloak whereas the new one only Quarkus dist of Keycloak.

Added a note about creds in CRs to make it more explicit.

design/keycloak.x/operator.md
vmuzikar: The two operator (old written in Go, new written in Java) won't be (most probably) possible to use at the same time. There'd be API conflicts (CRDs) and it wouldn't make much sense as the old one supports only Wildfly dist of Keycloak whereas the new one only Quarkus dist of Keycloak.

design/keycloak.x/operator.md
vmuzikar: @kfox1111 Added more details around the Ingress.

@hecsalazar In the initial version, we aim only at having simple opinionated Ingress/Route. We might consider extending this in the future.

design/keycloak.x/operator.md
vmuzikar: Elaborated on the DB.

vmuzikar:
@kfox1111 @hecsalazar Thank you for your feedback!

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+The Operator deployment is going to be performed leveraging OLM providing both CLI approach via `Subscription` objects for managing the operator installation, and UI in OpenShift. The Operator as such is going to be distributed as a container image.
+
+## Migration from the old operator
+No direct migration path. Generic migration steps will be documented.
+
+## Future considerations
+
+### Autonomous operator
+
+Our long-term vision includes making the operator autonomous to some extent, basically making it a [Level 5 operator](https://operatorframework.io/operator-capabilities/). It should be able to understand the operand's metrics and reflect them while automatically scaling and healing Keycloak deployment.
```
slaskawi, line 123: I already mentioned this the above - have a look at KEDA.

slaskawi (approved):
Great work! I really like the proposal!

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+**NOTE**
+
+The primary target of the operator is to make it easy to achieve production grade installations of Keycloak.X.
+
+---
+
+### Configuring Keycloak deployment
+
+The operator will use a CRD representing the Keycloak installation. The CRD will expose the following configuration options:
+* custom Keycloak image; default base image will be used if not specified
```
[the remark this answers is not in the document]
vmuzikar: I'm not sure this is something that makes sense for this operator (unless it's somehow "required" by OLM best practices which AFAIK it is not). With descoped/cluster-wide operators (which implies multi-tenancy) I believe it makes sense for the custom image to be primarily and explicitly set on the CR.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+* Git configuration to fetch static business objects configuration (see below)
```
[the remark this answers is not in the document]
vmuzikar: Good point. Added a chapter for configuring proxy.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+* manual horizontal and vertical scaling
```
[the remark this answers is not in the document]
vmuzikar: Thanks! Will look into it.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+* pod and node affinity, toleration
+* SSL config, truststore
```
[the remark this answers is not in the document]
vmuzikar: Thanks! Will look into it. However, we'd like to stay also vanilla K8s friendly so this would have to be an additional feature.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+
+Since most of the configuration options will come from the Keycloak.X distribution itself, the CRD will also expose appropriate fields for passing any distribution related options to the container, like database connection details (obviously without any credentials), SPIs configuration, etc.
+
+### Configuring Keycloak business objects using Kubernetes resources
+
+The new operator will expose two main ways of configuring business objects in Keycloak (e.g.: Realms, Roles, Clients, etc.) in addition to the built-in Dynamic configuration through the REST API/UI console:
+* Static file based configuration stored in Git to enable GitOps operations
+* Static configuration through Kubernetes API
```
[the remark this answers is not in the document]
vmuzikar: See the chapter below. ;)

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,123 @@
...
+
+It will be possible to store any credentials as Secrets in K8s leveraging [Keycloak Vault functionality](https://www.keycloak.org/docs/latest/server_admin/index.html#_vault-administration) where possible.
+
+It's purpose of the upcoming Store.X initiative to provide a full-fledged static configuration backend for Keycloak but there will be a mid-term preview to enable bulk imports at startup time leveraging the REST api.
+
+### Keycloak versions alignment
+
+The operator will have its release cycle aligned with Keycloak. Each operator version will support only one corresponding Keycloak version.
+
+### Upgrades
```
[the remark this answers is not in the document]
vmuzikar: As the operator version is tied to Keycloak version, there's no special treatment for downgrades. Operator downgrade would need to be done through OLM. But it should not be a problem thanks to rolling upgrades. I added a few more details on that.

design/keycloak.x/operator.md
vmuzikar: Thanks! Will look into it.

vmuzikar:
Hi @slaskawi! Long time no see. :) Thanks for the feedback!

design/keycloak.x/operator.md
kfox1111: it would only be a problem I think if you were trying to reuse the CRD's with the new operator as well as the old one.

Will the new one have its own API completely separate from the old one so the two can co'exist?

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+We will be leveraging Kubernetes volumes to act as "caches" for the [augmented/configured](https://quarkus.io/guides/reaugmentation) version of Keycloak.
+An initial POC to show the concept has been drafted here:
+https://github.com/andreaTP/poc-mutable-jar
+
+We will use Kubernetes volumes to cache the augmented version of the binaries.
+The artifacts in the kubernetes volume will be produced by an init-container and the operation is going to result in a noop in case the volume has already been populated by a compatible augmentation.
+
+### Connecting to a database
+
+A Postgres DB instance will have to be provisioned externally, it's not Keycloak Operator's responsibility to manage a database. The DB credentials will be stored as K8s Secrets.
```
kriswuollett, line 87: Would it be easier to fetch the DB connection parameters from Secret name/key pairs instead of making a Secret just for `keycloak-operator`? Same goes for the certificate for SSL. So if something like the [CrunchyData/postgres-operator](https://github.com/CrunchyData/postgres-operator) was being used, then something like the following had already been created:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: db1-pguser-db1
type: Opaque
data:
  dbname: db1
  host: db1-primary.database.svc
  jdbc-uri: jdbc:postgresql://db1-primary.database.svc:5432/db1?password=[...]&user=db1
  password: [...]
  port: 5432
  uri: postgresql://db1:[...]@db1-primary.database.svc:5432/db1
  user: db1-admin
  verifier: SCRAM-SHA-256$4096:[...]
```

If the secret wasn't in the right namespace, then something like [kubeops/config-syncer](https://github.com/kubeops/config-syncer) may also be in use. But once it is there, the CRD for Keycloak could look like:

```yaml
apiVersion: keycloak.org/v2
kind: Keycloak
metadata:
  name: example-keycloak
  labels:
    app: sso
spec:
  instances: 1
  externalDatabase:
    driver:
       value: postgres
    database:
      value: keycloak
   ssl:
     mode: verify-ca
     crtSecret:
       name: db1-cluster-cert
       key: ca.crt
    host:
       valueFromSecret:
         name: db1-pguser-db1
         key: host
    port:
       valueFromSecret:
         name: db1-pguser-db1
         key: port
    username:
       valueFromSecret:
         name: db1-pguser-db1
         key: user
    username:
       valueFromSecret:
         name: db1-pguser-db1
         key: password
```

Although mapping the values to a new Secrets isn't too difficult with Terraform, I'd guess it would be nice to minimize adding anything extra, as well as allowing for live updates to Secrets.

```suggestion
A Postgres DB instance will have to be provisioned externally, it's not Keycloak Operator's responsibility to manage a database. The DB connection parameters will be fetched from K8s Secret name/key pairs.
```

kriswuollett: For reference in the thread, currently a new Secret needs to be created with values you may already have in another Secret. Copied from [`external-db-secret.yaml` example](https://github.com/keycloak/keycloak-operator/blob/main/deploy/examples/keycloak/external-db-secret.yaml):

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: keycloak-db-secret
  namespace: keycloak
  labels:
    app: sso
data: # all values Base64
  POSTGRES_DATABASE:  <Database Name>
  POSTGRES_EXTERNAL_ADDRESS: <External Database  IP or URL (resolvable by K8s)>
  POSTGRES_EXTERNAL_PORT: <External  Database Port>
  POSTGRES_PASSWORD: <Database Password>
  POSTGRES_USERNAME:  <Database Username>
type: Opaque
```

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+---
+
+### Configuring Keycloak deployment
+
+The operator will use a CRD representing the Keycloak installation. The CRD will expose the following configuration options:
+* custom Keycloak image; default base image will be used if not specified
+* Git configuration to fetch static business objects configuration (see below)
+* manual horizontal and vertical scaling
+* pod and node affinity, toleration
+* SSL config, truststore
```
pedroigor, line 36: We support different proxy modes, some of them should not require TLS (e.g.: edge). But I guess we always want a secure channel between Keycloak and the ingress/route, even if running within the cluster network. Perhaps this requirement should be more explicit?

In addition to that, do we want to enable HTTP? Is this somewhat needed by Grafana (or has a complex set up) or any other client/service that needs to communicate with Keycloak?

I'm also missing exposing options to manage some key ingress/route settings. I think these are very common settings you usually want to change and have some control:

* connection limit
* timeout
* LB strategy

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+---
+**NOTE**
+
+The primary target of the operator is to make it easy to achieve production grade installations of Keycloak.X.
+
+---
+
+### Configuring Keycloak deployment
+
+The operator will use a CRD representing the Keycloak installation. The CRD will expose the following configuration options:
```
pedroigor, line 31: I'm missing the following configuration options:

* Clustering. Considering the current storage we might want to have options to set the number of `owners` for distributed caches, the number of entries, etc. Not sure how that would look like with the new store though, especially because the operator might need an Infinispan Server in place.
* JVM Settings. Would the operator provide specific options for setting specific JVM options (heap size, new size, gc, etc) or a single option to pass any JVM setting you want? I know this is something we should try to avoid as much as possible, but IMO this is a must (e.g.: capture gc logs)

design/keycloak.x/operator.md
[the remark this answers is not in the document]
pedroigor: You mentioned multi-tenancy. What is our main deployment model? 

* A single deployment serving multiple namespaces (and potentially multi-realms)
* A single deployment per namespace (single-realm)
* Both, up to the user

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Static configuration through Git
+
+The `Keycloak` CRD will enable defining a specific commit (identified by an hash) in a Git repository containing the static configuration in the form of JSON/YAML files. To update the configuration the user will simply change a commit hash in a `Keycloak` CR and the operator will roll out the new configuration to all pods.
+
+#### Static configuration through Kubernetes API
+
+The operator will leverage dedicated CRD(s), initially, there will be only one `Realm` CRD directly translated from Keycloak's [RealmRepresentation](https://github.com/keycloak/keycloak/blob/c7134fd5390d7c650b3dfd4bd2a2855157042271/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java). A Realm includes all subresources. As a result, it is going to be possible to configure every object in Keycloak through this CR even though for some of them it won't be recommended (e.g. Users). To implement this, the operator will simply translate the CRs to YAML files and mount them to Keycloak pods, again leveraging Store.X.
```
pedroigor, line 56: It is quite weird that we have the Realm CRD but we don't recommend configuring some key aspects like users. In practice, you are favoring managing users (and some others) through the admin console, right?

IMO, we should *provision* realms and then rely on the administration console. However, I see some space for improvement to enable some key capabilities/configuration when provisioning realms such as:

* Pre-configure for FAPI
* Pre-configured for a specific registration/authentication flow
* Brokering using social providers
* Any additional realm capability/configuration that otherwise the user would need to configure by himself.

We would have some realm templates that users can choose from to provision the realm.

Manage realms (fully) using CRs does not look user-friendly for me.

andreaTP: > We would have some realm templates that users can choose from to provision the realm.

this is a great idea!

design/keycloak.x/operator.md
DGuhr: +1 for this. Imo Caching/Clustering as well as the JVM environment is an infrastructure concern and thus should be handled by the operator. Not sure how Store.X fits in here, though.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Configuring Keycloak business objects using Kubernetes resources
+
+The new operator will expose two main ways of configuring business objects in Keycloak (e.g.: Realms, Roles, Clients, etc.) in addition to the built-in Dynamic configuration through the REST API/UI console:
+* Static file based configuration stored in Git to enable GitOps operations
+* Static configuration through Kubernetes API
+
+Static configuration will be strictly read-only, therefore two-way syncing is not going to be needed.
+
+Static configuration is going to provide an immutable and cloud native way for managing business objects in Keycloak that can be easily moved between environments (e.g. dev, stage, prod) in a predictable manner. This feature will leverage the new Store.X which enables federation of the configuration from multiple sources (static/dynamic) by re-architecting the storage layer.
```
DGuhr, line 48: Can you add a link to the appropriate Store.X capabilities here? I think this is rather vague and led me to "I want to know more".

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Static configuration through Git
+
+The `Keycloak` CRD will enable defining a specific commit (identified by an hash) in a Git repository containing the static configuration in the form of JSON/YAML files. To update the configuration the user will simply change a commit hash in a `Keycloak` CR and the operator will roll out the new configuration to all pods.
```
DGuhr, line 52: would be nice to see an example config

design/keycloak.x/operator.md
DGuhr: +1, i really like the idea of templates.

DGuhr, line 56: is it intended that the `RealmRepresentation` link doesn't reference the current main branch? If so, why? -> https://github.com/keycloak/keycloak/blob/main/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java

DGuhr, line 56: Can you elaborate a bit about `...though for some of them it won't be recommended (e.g. Users)...`? IMO Read-only means read-only, so it doesn't make much sense to recommend "do it automated for x, but manually for y".

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Upgrades
+
+In order to upgrade to a newer Keycloak version, the operator will be upgraded first to ensure full compatibility with the operand.
+
+If custom Keycloak image is not used, the operator will use a default base image. After the operator is upgraded, it automatically upgrades Keycloak too using a newer base image.
+
+In case a custom Keycloak image is used, the image will need to be rebuilt to perform the upgrade. This is not going to be operator's responsibility as building a custom image often requires a complex process incl. CI/CD pipelines. After the operator is upgraded, it won't manage any existing Keycloak instances until its custom image is manually rebuilt using the right Keycloak base image aligned with the operator and updated and the image coordinates are updated in the CR.
+
+Store.X will allow zero-downtime rolling upgrades (a Keycloak upgrade performed pod by pod) that will ensure that Keycloak cluster will remain available even when upgrade fails on one of the pods.
```
DGuhr, line 74: Just a thought: Might be that some people don't need the zero-downtime upgrade at all. Should we provide another path using the usual ways?

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+
+The operator will have its release cycle aligned with Keycloak. Each operator version will support only one corresponding Keycloak version.
+
+### Upgrades
+
+In order to upgrade to a newer Keycloak version, the operator will be upgraded first to ensure full compatibility with the operand.
+
+If custom Keycloak image is not used, the operator will use a default base image. After the operator is upgraded, it automatically upgrades Keycloak too using a newer base image.
+
+In case a custom Keycloak image is used, the image will need to be rebuilt to perform the upgrade. This is not going to be operator's responsibility as building a custom image often requires a complex process incl. CI/CD pipelines. After the operator is upgraded, it won't manage any existing Keycloak instances until its custom image is manually rebuilt using the right Keycloak base image aligned with the operator and updated and the image coordinates are updated in the CR.
```
DGuhr, line 72: Thinking about:
operator upgraded -> old kc unmanaged -> new image there but faulty, doesn't spin up -> so now keycloak is unmanagedly running in the k8s cluster. 

Which leads me to the question:
Is there an easy rollback strategy for the operator itself, to "go back to managed state until new image is ready and working"? I assume so (just re-releasing the old operator), but just to make sure we don't forget some error cases.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,128 @@
...
+Store.X will allow zero-downtime rolling upgrades (a Keycloak upgrade performed pod by pod) that will ensure that Keycloak cluster will remain available even when upgrade fails on one of the pods. 
+
+### Reaugmentation process in Kubernetes
+
+We will be leveraging Kubernetes volumes to act as "caches" for the [augmented/configured](https://quarkus.io/guides/reaugmentation) version of Keycloak.
+An initial POC to show the concept has been drafted here:
+https://github.com/andreaTP/poc-mutable-jar
+
+We will use Kubernetes volumes to cache the augmented version of the binaries.
+The artifacts in the kubernetes volume will be produced by an init-container and the operation is going to result in a noop in case the volume has already been populated by a compatible augmentation.
```
DGuhr, line 83: What happens when the reaugmentation fails for some reason? retry mechanism? messages, etc?

design/keycloak.x/operator.md
vmuzikar: The APIs should be separated, the new operator will use a different CRDs version.

design/keycloak.x/operator.md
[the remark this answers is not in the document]
vmuzikar: @pedroigor By deployment you mean operator or operand deployment?

For the operator, it'll be up to the user. Initially we'll probably support only namespaced mode (= single deployment per namespace) as we wait for proper descoped implementation in OLM.

design/keycloak.x/operator.md
vmuzikar: @kriswuollett Thanks for your feedback! Yes, we actually meant it basically as you described. But to be more explicit and clear I'll rewrite it a bit per your suggestion.

design/keycloak.x/operator.md
vmuzikar: We'll add more details around SSL.

The current plan is that the operator will provide only default Ingress setting. If you want to control Ingress and configure it, you can do it directly. You can even deploy a custom Ingress.

design/keycloak.x/operator.md
vmuzikar: I believe both areas are basically a dist config options and as such are configured as any other Quarkus dist option, that means env vars e.g. Is that correct? We have this already covered by:
> Since most of the configuration options will come from the Keycloak.X distribution itself, the CRD will also expose appropriate fields for passing any distribution related options to the container

design/keycloak.x/operator.md
vmuzikar: The templates functionality will basically have to come from the Store.X. Operator won't be actively involved in this. To my understanding, it will indeed work as you're suggesting. A "template" will be the static config (stored either in git or as a CR) while some config will be dynamic (modifiable through REST API / Admin Console) thanks to different store types aggregation. Correct @hmlnarik @mhajas ?

design/keycloak.x/operator.md
vmuzikar: @hmlnarik @mhajas What is the best "summary" doc on Store.X? :) There are several design proposals (some even unmerged yet) but is there anything that would give the whole picture?

design/keycloak.x/operator.md
andreaTP: Pretty much work in progress but this is an anticipation:
https://gist.github.com/andreaTP/9421dd3ffba0a039215139bbf7e6ce48

vmuzikar: We don't have the CRD fully defined yet and I believe this is an implementational detail. But it really should be simple. I can imagine a simple field in the CRD containing repo address and commit hash.

design/keycloak.x/operator.md
vmuzikar: It was current main at the time of writing it. ;) It's just my habit to keep it super-future proof. I can change it to main if you really believe it makes more sense. I don't have a strong opinion about it.

design/keycloak.x/operator.md
vmuzikar: That's a good point. I'll add a note about it.

design/keycloak.x/operator.md
andreaTP: +1 for keeping the permalink

design/keycloak.x/operator.md
vmuzikar: I can imagine a "reaugmentation fault" is the same as "pod won't start", i.e. readiness probes will be failing.

design/keycloak.x/operator.md
DGuhr: Yes, it is an implementation detail, but at least for me it helps a lot to wrap my head around the concept to see an example, as unfinished as it is :)

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,136 @@
...
+
+It will be possible to store any credentials as Secrets in K8s leveraging [Keycloak Vault functionality](https://www.keycloak.org/docs/latest/server_admin/index.html#_vault-administration) where possible.
+
+It's purpose of the upcoming Store.X initiative to provide a full-fledged static configuration backend for Keycloak but there will be a mid-term preview to enable bulk imports at startup time leveraging the REST api.
+
+### Keycloak versions alignment
+
+The operator will have its release cycle aligned with Keycloak. Each operator version will support only one corresponding Keycloak version.
```
benlangfeld, line 71: Surely you'll need to support at least two Keycloak versions per operator version to span the upgrade process?

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,136 @@
...
+We will be leveraging Kubernetes volumes to act as "caches" for the [augmented/configured](https://quarkus.io/guides/reaugmentation) version of Keycloak.
+An initial POC to show the concept has been drafted here:
+https://github.com/andreaTP/poc-mutable-jar
+
+We will use Kubernetes volumes to cache the augmented version of the binaries.
+The artifacts in the kubernetes volume will be produced by an init-container and the operation is going to result in a noop in case the volume has already been populated by a compatible augmentation.
+
+### Connecting to a database
+
+A Postgres DB instance will have to be provisioned externally, it's not Keycloak Operator's responsibility to manage a database. The DB credentials will be fetched from K8s Secrets.
```
benlangfeld, line 95: Is MySQL not going to be supported any more?

benlangfeld:
Is there an anticipated release schedule for this new operator in order to inform the decision (to wait or plan to re-implement) of those looking to roll out Keycloak on Kube in the near future?

design/keycloak.x/operator.md
stianst: Realm templates makes little sense to me. This should rather be addressed by improving the RealmRepresentation and making it easier to configure things than to have to rely on templates/examples.

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,136 @@
...
+
+It will be possible to store any credentials as Secrets in K8s leveraging [Keycloak Vault functionality](https://www.keycloak.org/docs/latest/server_admin/index.html#_vault-administration) where possible.
+
+It's purpose of the upcoming Store.X initiative to provide a full-fledged static configuration backend for Keycloak but there will be a mid-term preview to enable bulk imports at startup time leveraging the REST api.
+
+### Keycloak versions alignment
+
+The operator will have its release cycle aligned with Keycloak. Each operator version will support only one corresponding Keycloak version.
+
+### Upgrades
```
stianst, line 73: Would like to see some more on how upgrades will be performed, and how canary style upgrades will be achieved, as well as how rollback will be achieved.

stianst (changes requested):
We should break this up into separate discussions in:
https://github.com/keycloak/keycloak/discussions/categories/keycloak-x-new-kubernetes-operator

design/keycloak.x/operator.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Kubernetes compatibility matrix
+* OpenShift >=4.7
+* Vanilla Kubernetes >=1.20
+
+Other Kubernetes distributions are supported only in the best effort mode.
+
+## Distribution
+
+The Operator deployment is going to be performed leveraging OLM providing both CLI approach via `Subscription` objects for managing the operator installation, and UI in OpenShift. The Operator as such is going to be distributed as a container image.
```
sathieu, line 127: OLM is not used much except in OpenShift clusters, and Helm is widely used. What about providing an Helm chart, or at least plain manifests ?

benlangfeld: Indeed a dependency on OLM just to install the operator is very heavy and unrealistic for most consumers.

kfox1111: I use the olm on several vanilla k8s clusters. So it may be more common then you think. But, a helm option might be nice too.

benlangfeld: Sure, using OLM is possible, but by requiring it you're adding another hurdle that makes Keycloak less attractive and there is no documented reason why forcing use of OLM is necessary.

kfox1111: FYI, I'm just a user.

I have experienced some of the benefits of the olm when it comes to stuff like auto upgrading of things. That's still rather painful with helm. Having up to date instances is pretty important in a security sensitive service like keycloak...

Devs time is scarce. If all they have time for is to support one install system that works everywhere (k8s and openshift), though may be harder for some, I get that. Others can come along and implement other install methods if needed. I don't think this blocks that. This is just the description of the starting point I think.

benlangfeld: If the install method for the operator is any more complex than [this](https://github.com/keycloak/keycloak-operator/blob/main/deploy/operator.yaml), I think that counts as a bug in the operator rather than a design choice.

kfox1111: A. If thats true, no need for helm either then. Single static manifest. Done.
B. that link ignores all the rest of the stuff needed to install it properly, https://github.com/keycloak/keycloak-operator/tree/main/deploy
C. I really don't like operators that you just install the operator and it installs all the rest of itself. That gives it way too much power. Centralizing that in helm or olm is much preferable IMO.
D. Helm and crd's have a long history of not working well.

vmuzikar: Hello @sathieu @benlangfeld @kfox1111! Thank you for your feedback.

It will (and actually already [is](https://github.com/keycloak/keycloak/blob/main/operator/README.md)) of course possible to deploy the operator "manually" using kustomize. We won't be creating any artificial obstacles to use the operator this way, we actually want to make this experience as good as possible as it is the main way for using the operator during its development.

OLM is pretty much a standard now in the operator's world, with more than 200 operators in OperatorHub.io, which is not exclusive to OpenShift. So while OLM is not a hard dependency for the new operator, it is strongly recommended for the best experience.

We currently don't plan support for Helm.

DavidKittleSEL: The current keycloak-operator has the option to be installed either with our without OLM. I think forcing the added complexity of another management interface (OLM) makes very little sense. besides k8s, or openshift for that matter, do just fine managing operator lifecycle.

In my experience using k8s in production, most organizations are using CI/CD to manage application deployments. Forcing OLM in that scenario adds no value, and in fact detracts from one of the core tenants of reliability engineering, "Simplicity". I'm all for having the flexibility to use OLM for those that want/need it, but for most production environments forcing and additional management interface (OLM) just doesn't make sense.

vmuzikar: @DavidKittleSEL The new operator will use exactly the same approach as the old one. Primarily distributed through OLM with the possibility to deploy the operator manually without it.

diversit:
Have there been some actual work on the new operator already or is it al just talk until now?
We're about to move our Keycloak deployments to K8S. I already have a setup with the 'old' operator, but obviously, I want to use the new Quarkus based Keycloak and not the old Wildfly based installs. I had assumed the v17 operator would use the v17 Quarkus based Keycloak images, but after an afternoon of setting up a new Kustomized based config I just noticed it's still using the old legacy Keycloak. 😞

andreaTP:
Hi @diversit , the new Keycloak operator for the Quarkus distribution will be available in preview with Keycloak 18 ref:
https://www.keycloak.org/2022/03/releases

Submoon:
Hi! I'm curious about the operator too because I had the same experience as @diversit 😄 

If the first tech preview version is supposed to be released with KC 18, when could we reasonably expect a stable "production-ready" operator?

vmuzikar:
@Submoon Unfortunately, I can't provide any details on when the new operator will reach "stable" release phase.

PR #335, opened 2022-01-21
Detailing downgrading of schema version 

ahus1:
Scenario: write in mixed environment or downgrades.

This is following up on the discussion https://github.com/keycloak/keycloak/discussions/9640


sguilhen (approved):
I personally think the suggested changes properly convey what was discussed and to me they make the document clearer.

PR #340, opened 2022-03-04
Update client-secret-rotation.md

marcelomrwin:
Update with the implemented flow

### DIFF
--- a/design/client-secret-rotation.md
+++ b/design/client-secret-rotation.md
@@ -84,12 +84,13 @@ This approach allows a global control over client secrets within a realm, and ha
 backwards compatible as there would be no change to the default behaviour. It also fits better into dynamic client
 registration than the previous proposal.
 
-Admin notification events
+Notification events
 -------------------------
 
-Throw an administrative events related to rotation and expiration of secrets.
+Throw events related to rotation and expiration of secrets.
 - When a client rotates a secret. Client ID and Client Name / Description.
 - When trying to authenticate using an expired rotated secret.
 - When a _percentage_ or _fixed amount_ remains to expire the secret. At this point we can add a new setting to the secret rotation.
   - Launch notification events when 10% of the time remains
   - Launch notification events after X seconds
+- Make use of the client's contacts field to notify via email. (Need to implement support for this field in client settings). 
--- a/design/client-secret-rotation.md
+++ b/design/client-secret-rotation.md
@@ -2,9 +2,8 @@ The idea is to introduce a client secret policy part of client policies, similar
 Initially it would not be extensible, but would be just a simple key/value configuration.
 
 We'd have the following options in client secret policy to cover secret rotation. The options `Secret expiration`, `Rotated secret expiration` 
-and `Remaining expiration for rotation during update` are applicable just if `Secret rotation enabled` is ON:
+and `Remaining expiration for rotation during update` are applicable just if the policy is applied to the client:
 
-- **Secret rotation enabled: [boolean]**
 - **Secret expiration: [seconds]** - When the secret is rotated, this is the expiration of time of the new secret
 - **Rotated secret expiration: [seconds]** - When secret is rotated, this is the remaining expiration time for the old secret.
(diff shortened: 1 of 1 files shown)


design/client-secret-rotation.md
```diff
@@ -78,3 +83,13 @@ The changes needed in the Admin UI for a client would be:
 This approach allows a global control over client secrets within a realm, and has less changes to the APIs, and is fully
 backwards compatible as there would be no change to the default behaviour. It also fits better into dynamic client
 registration than the previous proposal.
+
+Admin notification events
+-------------------------
+
+Throw an administrative events related to rotation and expiration of secrets.
+- When a client rotates a secret. Client ID and Client Name / Description.
+- When trying to authenticate using an expired rotated secret.
```
mposolda, line 92: Currently when authentication with invalid client secret is done, the "normal" event is triggered, but not the "admin event" . AFAIK it is the normal error event with error like "invalid client authentication" . I guess this should not be different when authentication with expired rotated secret is done?

Maybe we can add some "detail" to the event that invalid client authentication was tried with expired "rotated" secret. But IMO it should be same event like during failed client authentication with expired "regular" client secret.

design/client-secret-rotation.md
```diff
@@ -78,3 +83,13 @@ The changes needed in the Admin UI for a client would be:
...
+- When a _percentage_ or _fixed amount_ remains to expire the secret. At this point we can add a new setting to the secret rotation.
```
mposolda, line 93: I recall that we earlier discussed together with Stian to add support to the "contacts" as the configuration option of the client. This option is also supported in the OIDC specification https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata .

This will allow that when client authentication is done and there is too small amount of time to expire the client secret (Maybe specified by "Remaining rotation period" or some other way), we can send the email to the client administrator, who was specified in the clients "contacts" . Is it what you meant?

mposolda (changes requested):
@marcelomrwin Thanks for sending the PR. I've added few comments to the text below. Can you please check it?

design/client-secret-rotation.md
marcelomrwin: Yes, it is a normal event. I was wrong when I mentioned administrative event. My suggestion is to add customer id and their description/name as detail.

mposolda: Hmm... I am not sure what do you mean by "customer id" ? Didn't you mean "client id" ? :-)

design/client-secret-rotation.md
mposolda: But I think that this is probably something to consider as a follow-up (Milestone 2). We probably don't manage to add this to RHSSO 7.6 though...

marcelomrwin: I confess I didn't remember this detail but it's exactly the point I want to get to. The initial idea was to be reactive by launching events from a certain point to notify administrators that possibly the secrets would expire within a time window. Regarding _"Remaining rotation period"_ this parameter was created to force rotation of dynamic clients. In this case for dynamic clients there will be no notification but the rotation directly. I think a new field is more correct because that way we can use it for common and dynamic clients.

Returning to the topic of **_contacts_**, I think it makes much more sense because we will be proactive. This should be a new activity that would serve as a prerequisite for the notifications theme, do you agree?

Certainly, we must consider the frequency of sending these emails, I believe that sending an email with each authentication that is about to expire makes the emails quickly considered SPAM.

design/client-secret-rotation.md
marcelomrwin: Yeah!! sorry again. That's what I meant, not customer, but **client**

design/client-secret-rotation.md
mposolda: +1 for consider the frequency of sent emails somehow. I am not sure about re-using the existing option or use the existing one...

It will be probably good to start dedicated discussion about this - maybe it even deserves to do the presentation on keycloak-weekly call for this after RHSSO 7.6 release :-) We can discuss more details about this...

mposolda (approved):

PR #341, opened 2022-03-26
docs: correct url path to /auth/device

Jeff-Tian:
/device/auth is wrong which cause the error for example: 

```
{
    "error": "RESTEASY003210: Could not find resource for full path: https://keycloak.jiwai.win/auth/realms/UniHeart/protocol/openid-connect/device/auth"
}
```

### DIFF
--- a/design/oauth2-device-authorization-grant.md
+++ b/design/oauth2-device-authorization-grant.md
@@ -99,7 +99,7 @@ Send POST request to the `Device Authorization Endpoint` like below.
 ```
 curl -X POST \
     -d "client_id=foo" \
-    "http://localhost:8080/realms/test/protocol/openid-connect/device/auth"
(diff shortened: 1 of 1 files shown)


stianst:
These are old design documents and not documentation. We're not actively maintaining these after implementation is completed.

stianst:
Rather than updating old and abandoned design documents we should mark these as archived, with a warning that they are likely to not include accurate information.

Another thing here is that it doesn't seem like our official documentation actually has any mention of device flow, etc. @mposolda is that correct or am I just looking in the wrong place?

stianst (changes requested):
See previous comments

stianst:
Guessing the reason why a lot of people are looking at this file is it pops up as the #1 result on a Google search for `keycloak device authorization grant`, and there's no mention of the proper documentation.

PR #353, opened 2023-12-14
Update specifications

tnorimat:
add FAPI 2.0 security profile, FAPI 2.0 message singing, RFC 9207, and RFC 9449 in keycloak 23.


tnorimat:
@mposolda I updated specification section. Could you or your team check the update?

